PR #9, opened 2015-03-24
Extend credintials.json from environment

### LINKED ISSUE #8: Use ENV variables
The app is hard to deploy and host because you have to deal with a JSON config file that you don't want checked in but needs to be on the server. It would make sense for the config object to try to read from ENV and fall back to defaults. This would make it easier to deal with and deploy.

matejkramny:
Resolves #8

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -1,13 +1,36 @@
-var fs = require('fs');
+var fs = require('fs'),
+	env = process.env;
 
 try {
-	var credentials = require('./credentials.json')
+	var credentials = require('./credentials.json');
 } catch (e) {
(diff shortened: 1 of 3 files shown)


lib/config.js
```diff
@@ -1,14 +1,37 @@
...
+	exports.credentials.smtp.user = env.SMTP_USER;
+}
+if (env.SMTP_PASS) {
+	exports.credentials.smtp.pass = env.SMTP_PASS;
+}
+if (env.MONGODB_URL) {
+	exports.credentials.db = env.MONGODB_URL;
+}
+
+console.log(exports.credentials)
```
Echobob: Is this console log necessary?

PR #10, opened 2015-03-24
fix vagrant install to run out of the box

benwyrosdick:
Get the vagrant deploy working out of the box:

* fixed copy of the credentials in playbook
* run with `node` command instead of `grunt`


matejkramny:
:+1:

Thanks

PR #15, opened 2015-03-31
More complete ENV migration

bibby:
removed credentials in favor of dockerfile ENV vars with defaults. docker-compose example file.


lib/config.js
```diff
@@ -1,39 +1,30 @@
...
-if (env.MONGODB_URL) {
-	exports.credentials.db = env.MONGODB_URL;
-}
+};
 
 exports.version = require('../package.json').version;
 exports.hash = 'dirty';
 exports.production = process.env.NODE_ENV == "production";
-exports.port = process.env.PORT || credentials.port;
+exports.port = env.SSL ? 443 : 80;
```
bibby: should be `env.USE_SSL`!

bibby:
I recognize now that my changes are not friendly to Vagrant users. Sorry for that. Hopefully we can turn this into a happy medium.

matejkramny:
Hmm. Maybe change this PR so the config file is still supported, while keeping the names of the env variables as you've got them.

Eg config file would be entirely optional if environment variables are present.

That seems like a good compromise between both worlds. What do you think?

bibby:
I honestly don't know how Vagrant likes to configure its applications, but I should put the config file back.
Recognizing Docker could be as sneaky/easy as:
```
ENV IS_DOCKER 1
```
..and swapping behavior based on that.

matejkramny:
I do not agree with keeping all default env variables inside the dockerfile because the app is also deployed outside of docker.

Running the app outside of docker therefore will not work.

In order for to merge this PR:

- We use tabs through the project. Please correct that on relevant lines (eg https://github.com/ansible-semaphore/semaphore/pull/15/files#diff-c39f41208533f29db9e4150b6efd2b68R31)
- Move environment variables to be defaulted inside config.js
- Make sure app can be run w/o docker
- See if fig.yml can be merged with docker-compose.yml
- Rebase onto master

bibby:
That's ok, bro. I took a docker pattern and over-extended it.

matejkramny:
no probs. Thanks for your time.

PR #27, opened 2015-06-07
Write a key pair by identity name

rmessner:
Avoid to rewrite all the time the default private/public key of the user running semaphore by creating a keypair by identity.

### DIFF
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -83,7 +83,7 @@ function installHostKeys (task, playbook, done) {
 	// Install the private key
 	playbookOutputHandler.call(task, "Updating SSH Keys\n");
 
-	var location = '/root/.ssh/id_rsa';
+	var location = '/root/.ssh/' + playbook.identity.name;
 	fs.mkdir('/root/.ssh', 448, function() {
 		async.parallel([
 			function (done) {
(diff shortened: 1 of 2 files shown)


matejkramny:
Hi, thanks!

What if `playbook.identity.name` contains a space or special characters that break the commands, eg causing invalid files?

I feel like there should be another field for if its a private key, what the name of it would be.

Also, semaphore is supposed to run in a sandbox, the key is saved in `~/.ssh/id_rsa` to save complications coming from git (which you've solved in the commit).

They could now be stored in a special dir eg. `~/.semaphore`.

Just a few ideas. Otherwise, thanks for your work!

acca301:
It's a little bit more complex - the whole pullGit.sh/ssh_wrapper.sh concept needs rework.

- the identity may be not RSA, but say, ED25519
- Git repo may be given in a URL form like "ssh://host:port/directory/repo". ssh_wrapper.sh chokes up on that

It sounds like **playbook.identity.name** should be introduced indeed, but with a different than the pull request semantics and pullGit.sh to be retired.

Still scratching my head, can we discuss that?

matejkramny:
@acca301 create a separate issue unless it directly involves this PR.

rmessner:
I closed it because i have no more time to spend on it.

PR #40, opened 2015-10-08
Replace root home with home of the semaphore user

### LINKED ISSUE #39: Run semaphore as normal user
Hi,

I really like semaphore, but the problem is that I don't run my infrastructure in docker and I don't want to run semaphore as `root` user. 

It would be great if we could run semaphore as a normal user eg. `semaphore` .

brodul:
Hi,

I replaced the '/root/' with the actual home of the running user. So one can start semaphore as a normal user. 

Fixes #39

### DIFF
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -6,6 +6,9 @@ var config = require('./config'),
 	models = require('./models'),
 	app = require('./app');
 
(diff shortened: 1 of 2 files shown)


Echobob:
:+1:

PR #46, opened 2015-10-13
Gives the ability to Edit a Job and Implements extra-vars

nodefourtytwo:
You can edit a job from the job list of a playbook and add some extra vars to be used when running the playbook.


QIvan:
Very useful feature!

matejkramny:
Thanks for your work. Very good

PR #50, opened 2015-10-22
If users in db don't create default user (fixes #48)

### LINKED ISSUE #48: Default admin user gets recreated even if deleted from panel
Default admin user gets recreated even if deleted from panel

https://github.com/ansible-semaphore/semaphore/blob/master/lib/config.js#L66

brodul:
If there are any users in the DB skip the creation of the default admin.

When user starts the app from scratch the default admin is created. If the user adds another admin and deletes the default user,
the default admin will not be created.
If there is the case that the user deletes all admins (including the one he is using right now) the default admin is created again.

This fixes bug #48 which can be a security issue for open deploys of semaphore application.

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -61,12 +61,11 @@ exports.init = function () {
 	var models = require('./models');
 
 	models.User.findOne({
-		email: 'admin@semaphore.local'
-	}).exec(function (err, admin) {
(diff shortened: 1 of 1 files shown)


matejkramny:
:+1: Thanks!

PR #63, opened 2016-02-08
Adding host variables to db

carlos-alarcon:
Based on https://github.com/ansible-semaphore/semaphore/pull/26
I created a 'vars' field in the form so the would be kept in DB
and used in the inventory creation


matejkramny:
:+1: Thanks!

PR #70, opened 2016-03-23
Added `tini' to reap zombies - a well-known issue with PID 1 process.

acca301:
Every job run runner.js leaves a "defunct" pullGit.sh. Please see https://blog.phusion.nl/2015/01/20/docker-and-the-pid-1-zombie-reaping-problem/ for details.


matejkramny:
Thanks :+1:

PR #89, opened 2016-05-16
Inject ansible-playbook executable options using job definition

samarjit:
Similar to extra_vars, I created exec_options which will be added to ansible_playbook  [options]. Currently I am using this to view details logs, but it can be generic and all ansible_playbook options can  be sent over. Please include this in master if you find it useful. 

Changes: I changed Job model, job edit template, jobs listing template, and save() and add() methods where Job gets persisted. Also changed runner.js to apply the changes while invoking executable.

### DIFF
--- a/lib/models/Job.js
+++ b/lib/models/Job.js
@@ -12,6 +12,7 @@ var schema = mongoose.Schema({
 	},
 	name: String,
 	play_file: String, //x.yml
+	exec_options: String,
 	extra_vars: String,
 	use_vault: Boolean
 });
--- a/lib/routes/job/job.js
+++ b/lib/routes/job/job.js
@@ -50,6 +50,7 @@ function remove (req, res) {
 function save (req, res) {
 	req.job.name = req.body.name;
 	req.job.play_file = req.body.play_file;
+	req.job.exec_options = req.body.exec_options;
 	req.job.extra_vars = req.body.extra_vars;
 
 	req.job.save();
--- a/lib/routes/job/jobs.js
+++ b/lib/routes/job/jobs.js
@@ -37,6 +37,7 @@ function add (req, res) {
 		playbook: req.playbook._id,
 		name: req.body.name,
 		play_file: req.body.play_file,
+		exec_options: req.body.exec_options,
 		extra_vars: req.body.extra_vars,
 		use_vault: req.body.use_vault
 	})
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -209,6 +209,14 @@ function playTheBook (task, playbook, done) {
 		}
 	}
 
+	if(task.job.exec_options.length>0){
+		var opts = task.job.exec_options.split(" ");
+
+		for (var i=0; i < opts.length; i++){
+			args.push(opts[i]);
+		}
+	}
+
 	console.log(args);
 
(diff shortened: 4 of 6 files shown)


matejkramny:
Hi @samarjit,

Your work is welcome but before it is merged a couple of things need to be done:
- Rebase the work and squash repeating commits
- Lines 211 onwards should be formatted according to https://google.github.io/styleguide/javascriptguide.xml?showone=Code_formatting#Code_formatting (lines here: https://github.com/ansible-semaphore/semaphore/pull/89/files#diff-b3b53682a18f203ac8d29b0e277cad26R211).

We also have v2 which is due to be released. It has been rewritten (mostly on the backend) and you should direct your contributions there.

samarjit:
Hi Matej,

I formatted the code and squashed the commits into one commit 8b5a5958034c3c3ca012e91dd35a9667a8b93aa6. 

Hope this time it helps you.

I would have applied the same changes to V2 as well. But I have never used golang. The new project structure also seems to be complete rewrite. Hope to see this version soon. 

Just curious. If you don't mind can you share the justification for moving to golang?

matejkramny:
The code backend was a bit of a mess..

From the v2 milestone:

> The goal of this is to create a super-simple interface which is capable of configuring majority of ansible, > and for those who have 'extra needs' there will be option to configure the command line params per task.
> 
> The need for v2 is also to improve deployment & maintainability of the project.

It extends on the functionality of v1 and extends with extra features for power users. One of the merits of this is single-file deploy and upgrade capabilities. I think users of semaphore will find it much easier to use & install.

matejkramny:
We can merge this into v1 before it is closed off. v2 already has this feature baked in :)

samarjit:
Hi Matej,

Thanks for merging. I was curious about what favored the language selection "go" instead of java, groovy, ruby or python?

Regards,
Samarjit

matejkramny:
My favourite toy :P

Jokes aside, here are my pros & cons:

- Very complete network library (supporting http/2)
- Compiles to a single binary, easy to distribute, cross-compile. I do lots of devops things, and this is my favourite point
- It is strict and forces you to write in a certain way. This is good because when you read other people's code, it is easier to understand.
- No NPM nonsense!
- Speed, easy multi-threaded processing
- Well designed i think, but takes a bit of time to understand. Integrates well with git
- Self-documenting code / testing! (though i haven't done much of it here)
- I think Go should have a `math.Round` function.. It doesn't have one currently, you have to be a bit inventive which leads you to bugs such as with negative numbers

I felt Go would be utilised best with semaphore as it is a HTTP API. Sadly one mistake whilst making v2 is choosing MySQL over MongoDB. MongoDB would've been a better choice, it is too late now though

samarjit:
Great thanks for enlightening. I assumed you were already facing maintenance problems with loose typing of javascript. NPM is a problem IED might replace it.   I was considering the second point above if it had been a jvm language its portable right away no cross compilation, but jre needs to be installed in every server though. But I guess, your point is have single complete executable and just run it, why bother with installing hell lot of libraries in all the systems. I will give it a 'Go'.

matejkramny:
I looked up IED but tbh doesn't solve anything other than performance..

PR #173, opened 2016-08-01
Fix DB migration - alter table

### LINKED ISSUE #172: Error in semaphore Installation
i am not able to install semaphore on Ubuntu 14

i got an error while running
# semaphore -setup

Executing migration v1.1.0 (at 2016-07-26 21:36:46.01249632 -0700 PDT)...
[1/1]
ERR! Query: /\* update precision of output logs - also prevents confusion when sorting the table to guarantee it is in the same order */
alter table task__output change time time datetime(6) not null;

Rolling back v1.1.0 (time: 2016-07-26 21:36:46.013665314 -0700 PDT)...
Rollback SQL doesn't exist.

Database migrations failed!
Error 1064: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '(6) not null' at line 2

Please help

willgarcia:
Closes #172

### DIFF
--- a/db/migrations/v1.1.0.sql
+++ b/db/migrations/v1.1.0.sql
@@ -1,2 +1,2 @@
 /* update precision of output logs - also prevents confusion when sorting the table to guarantee it is in the same order */
-alter table task__output change `time` `time` datetime(6) not null;
+alter table task__output change `time` `time` datetime not null;


matejkramny:
Precision is required because logs can be sent within a millisecond of each other. This would confuse the order of the log messages.

If the precision needs to be removed, another column needs to be introduced. The datetime could be split up, nanoseconds in another column.

PR #207, opened 2016-10-18
Erorr -> Error

jahantech:
Incorrect Spelling.


matejkramny:
I wonder where that sloppy mistake came from. Thanks for fixing!

PR #209, opened 2016-10-25
add ssh agent to docker image

brunowoo:


matejkramny:
What's up?

It should tag `latest`

PR #227, opened 2016-11-17
add support for git repo tags or branches 

knsr:
Correct issue [119](https://github.com/ansible-semaphore/semaphore/issues/119).

With this, you just have to specify the tag or branch at the end of url in the playbook repository definition.
`git://github.com:userspace/reponame.git#tag|branch`

_Full disclosure : I am not a golang dev ;)_


matejkramny:
Would it be preferred to put the branch as a tag or an extra field in the UI?

PR #228, opened 2016-11-17
Fix for migrations of the new "Alias" Field.

gcavalcante8808:
Fixes the migration to field "Alias" on the project_template which was introduced by https://github.com/ansible-semaphore/semaphore/commit/ed46338da541e1d81707cdf1cbe5d5f9f11310bd.

Previously the Sql file exists but wasn't referenced on bindata.go and versionHistory.go.


matejkramny:
Thanks

PR #241, opened 2016-12-22
fix error api-doc route

pianzide1117:
/project/{project_id}/template   ==>   /project/{project_id}/templates


Echobob:
Thanks @pianzide1117 👍

Issue #255, opened 2017-02-07
Synchronize module can't find rsync executable


kpashka:
I'm using Ubuntu 16.04, Ansible 2.2.1.0, both hosts have `rsync` installed (in `/usr/local/bin`), 100% sure. One of the playbooks has `synchronize` module, and, for some reason, Semaphore can't find the executable:

```
21:44:13: fatal: [10.240.0.3]: FAILED! => {
21:44:13:     "changed": false, 
21:44:13:     "failed": true, 
21:44:13:     "invocation": {
21:44:13:         "module_args": {
21:44:13:             "_local_rsync_path": "rsync", 
21:44:13:             "_substitute_controller": false, 
21:44:13:             "archive": true, 
21:44:13:             "checksum": false, 
21:44:13:             "compress": true, 
21:44:13:             "copy_links": null, 
21:44:13:             "delete": true, 
21:44:13:             "dest": "10.240.0.3:/***", 
21:44:13:             "dest_port": null, 
21:44:13:             "dirs": false, 
21:44:13:             "existing_only": false, 
21:44:13:             "group": null, 
21:44:13:             "links": null, 
21:44:13:             "mode": "push", 
21:44:13:             "owner": null, 
21:44:13:             "partial": false, 
21:44:13:             "perms": null, 
21:44:13:             "private_key": "/tmp/semaphore/access_key_1", 
21:44:13:             "recursive": true, 
21:44:13:             "rsync_opts": null, 
21:44:13:             "rsync_path": "\"sudo rsync\"", 
21:44:13:             "rsync_timeout": 0, 
21:44:13:             "set_remote_user": true, 
21:44:13:             "src": "/tmp/semaphore/repository_1/***/***", 
21:44:13:             "ssh_args": null, 
21:44:13:             "times": null, 
21:44:13:             "verify_host": false
21:44:13:         }
21:44:13:     }, 
21:44:13:     "msg": "Failed to find required executable rsync"
21:44:13: }
```

matejkramny:
Interesting, could you try running this w/o semaphore to rule it out?

It might be the setup of the box ansible is ran on or a permission issue perhaps.

kpashka:
@matejkramny I was also running Semaphore locally on my Mac, had exactly the same issue. This specific playbook was working smoothly for like decades, but under semaphore something got wrong. Gonna try to debug it deeper today

matejkramny:
I have a feeling it might be related to the `$PATH` variable which:

[...]

PR #256, opened 2017-02-08
Pass OS environment variables to Ansible

kpashka:
Related issue #255

### DIFF
--- a/make.sh
+++ b/make.sh
@@ -38,7 +38,7 @@ cd -
 
 echo "Adding bindata"
 
-go-bindata $BINDATA_ARGS config.json db/migrations/ $(find ./public -type d -print)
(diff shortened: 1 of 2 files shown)


make.sh
```diff
@@ -38,7 +38,7 @@ cd -
 
 echo "Adding bindata"
 
-go-bindata $BINDATA_ARGS config.json db/migrations/ $(find ./public -type d -print)
+go-bindata $BINDATA_ARGS config.json db/migrations/ $(find public/* -type d -print)
```
kpashka: Also added this fix, because paths with trailing slashes are being aborted with 404 for some reason (was struggling with building & running locally):
https://github.com/ansible-semaphore/semaphore/blob/master/api/router.go#L115

matejkramny: 👍

matejkramny:
Very good, thanks.

PR #268, opened 2017-02-21
fix api breaking of google/go-github

### LINKED ISSUE #267: Build fails because of API breaking of google/go-github
Currently the build of semaphore fails with the following error: 

```
--> linux/amd64 error: exit status 2                                                                                   
Stderr: # github.com/ansible-semaphore/semaphore/util                                                                  
../util/upgrade.go:99: not enough arguments in call to gh.Repositories.ListReleases
```

This is because the used go lib `google/go-github` broke their API with a recent commit to master:  
https://github.com/google/go-github/commit/23d6cb9cacb5aa314e93d600fe20a48496a718d4

nightvisi0n:
google/go-github broke the api of their functions with google/go-github@23d6cb9
This PR adds the context `context.TODO()` to the call of `gh.Repositories.ListReleases` which is meant to be used "when it's unclear which Context to use or it is not yet available (because the surrounding function has not yet been extended to accept a Context parameter)" (citing https://godoc.org/context)  
I chose this way because AFAIK we don't have much use of the [`context` package](https://golang.org/pkg/context/).  

fixes #267

### DIFF
--- a/util/upgrade.go
+++ b/util/upgrade.go
@@ -10,6 +10,7 @@ import (
 	"path/filepath"
 	"runtime"
 	"time"
+	"context"
 
 	"github.com/google/go-github/github"
 )
@@ -96,7 +97,7 @@ func findAsset(release *github.RepositoryRelease) *github.ReleaseAsset {
 func CheckUpdate(version string) error {
 	// fetch releases
 	gh := github.NewClient(nil)
-	releases, _, err := gh.Repositories.ListReleases("ansible-semaphore", "semaphore", nil)
+	releases, _, err := gh.Repositories.ListReleases(context.TODO(), "ansible-semaphore", "semaphore", nil)
 	if err != nil {
 		return err
 	}


nightvisi0n:
@matejkramny I would propose to do at least a hotfix release (2.1.1) after this is merged in order to have a release that builds.  
The hotfix release could just contain this commit (cherry-pick) if you don't yet want to release all the other commits since https://github.com/ansible-semaphore/semaphore/commit/8d2e741813ffbf346e342338f918281273cfe59e.

matejkramny:
Thanks Julian.

I'll produce a release in the next 48hrs

nightvisi0n:
> I'll produce a release in the next 48hrs

Thank you! This will be really helpful for everyone that uses semaphore by builting the latest release from source. :slightly_smiling_face:

matejkramny:
Sure, I might try to automate it as it's a little tedious.. There's also another PR which should be merged

PR #294, opened 2017-03-12
Added primary keys to the existing tables

### LINKED ISSUE #293: Add MySQL primary keys to the tables
This would:
* Improve MySQL index performance
* Fixes WSREP issue on Percona XtraDB MySQL about `un-recommended use of DML command on a table without explicit primary key`, when running Percona in `pxc_strict_mode` `MASTER` or `ENFORCING` (full error message see below).
```
WSREP: Percona-XtraDB-Cluster doesn't recommend use of DML command on a table (semaphore.task__output) without an explicit primary key with pxc_strict_mode = PERMISSIVE
```

galexrt:
This adds primary keys to the database tables.

This closes #293

### DIFF
--- a/db/migrations/v2.2.1.sql
+++ b/db/migrations/v2.2.1.sql
@@ -1,0 +1,27 @@
+ALTER TABLE `user` ADD PRIMARY KEY(id)
+
+ALTER TABLE `project` ADD PRIMARY KEY(id)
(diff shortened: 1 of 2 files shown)


matejkramny:
Thanks for fixing this 👍

PR #297, opened 2017-03-14
Add copy button to "Task Templates" page

### LINKED ISSUE #259: Allow copying a task templates
In our infrastructure, we use one big playbook which can install/update the entire infrastructure. Separate parts of this playbook we launch with various tags. 
So, in Semaphore paradigm we using almost equal task templates (the same Playbook Name, SSH Key, Inventory, Playbook Repository and Environment, but different Extra CLI Arguments (["--tags=app1"], ["--tags=app2"], ["--tags=app3"])). For almost equal templates Clone/Copy button will save many time.

laeshiny:
Fix #259

### DIFF
--- a/public/js/controllers/projects/templates.js
+++ b/public/js/controllers/projects/templates.js
@@ -126,6 +126,30 @@ define(['controllers/projects/taskRunner'], function () {
 			})
 		}
 
+		$scope.copy = function (template) {
+            var tpl = angular.copy(template);
+            tpl.id = null;
+
+		    var scope = $rootScope.$new();
+			scope.tpl = tpl;
+			scope.keys = $scope.sshKeys;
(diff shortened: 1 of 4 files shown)


public/js/controllers/projects/templates.js
```diff
@@ -126,6 +126,30 @@ define(['controllers/projects/taskRunner'], function () {
 			})
 		}
 
+		$scope.copy = function (template) {
+            var tpl = angular.copy(template);
```
matejkramny: please reformat to the file (can see indentation differences here) :)

laeshiny: oh... I'm sorry and thank you for your review. I correct to reformat :)

PR #299, opened 2017-03-14
Fix the primary key creation queries in v2.2.1 migration

galexrt:
* Add id column to task__output table (fix the queries for it)
* Print error message in case of database error(s)

I tried to fix the `util` test but failed in doing so..

***

Should the PR target the `develop` or `master` branch ? (This PR is targeted at the `develop` branch)

### DIFF
--- a/db/migrations/v2.2.1.sql
+++ b/db/migrations/v2.2.1.sql
@@ -1,27 +1,5 @@
-ALTER TABLE `user` ADD PRIMARY KEY(id)
+ALTER TABLE `project__user` ADD PRIMARY KEY(project_id);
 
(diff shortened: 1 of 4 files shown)


db/migrations/v2.2.1.sql
```diff
@@ -1,27 +1,5 @@
-ALTER TABLE `user` ADD PRIMARY KEY(id)
+ALTER TABLE `project__user` ADD PRIMARY KEY(project_id);
```
matejkramny: this primary key makes it impossible to add > 1 user to a project since the key structure is a unique key `user_id`+`project_id`

PR #300, opened 2017-03-14
Improve the docker entrypoint and dockerfile

galexrt:
* Improve the Dockerfile build speed and space by flattening all to one layer
* Added env var in Dockerfile for version and arch of semaphore


matejkramny:
Thanks for your time @galexrt

PR #366, opened 2017-05-29
Allow concurrency for tasks that does not collide. Fixes #365

### LINKED ISSUE #365: Concurrent deploys
In our use case, we make slightly below 10 000 deploys each month. When using semaphore, we deploy to prod and test servers in different projects. We currently have to wait to make deploys to production servers if someone is deploying to a test server or vice versa. This becomes a problem, especially when the job takes long time because it triggers server restarts or similar, halting the development process. It would also be nice to be able to deploy to different test servers simultaneously if one knows that there's no collision for the hosts.

Natanande:
Two different concurrency modes are implemented, and is enabled by
setting "concurrency_mode" in the config file to either "project" or "node".

When "project" concurrency is enabled, tasks will run in parallel if and
only if they do not share the same project id, with no regard to the
nodes/hosts that are affected.

When "node" concurrency is enabled, a task will run in parallel if and
only if the hosts affected by tasks already running does not intersect
with the hosts that would be affected by the task in question.

If "concurrency_mode" is not specified, no task will start before the
previous one has finished.

The collision check is based on the output from the "--list-hosts"
argument to ansible, which uses the hosts specified in the inventory.
Thus, if two different hostnames are used that points to the same node,
such as "127.0.0.1" and "localhost", there will be no collision and two
tasks may connect to the same node concurrently. If this behaviour is
not desired, one should make sure to not include aliases for their hosts
in their inventories when enabling concurrency mode.

To restrict the amount of parallel tasks that runs at the same time, one
can add the "max_parallel_tasks" to the config file. This defaults to a
humble 10 if not specified.

### DIFF
--- a/api/tasks/pool.go
+++ b/api/tasks/pool.go
@@ -3,46 +3,112 @@ package tasks
 import (
 	"fmt"
 	"time"
+
+	"github.com/ansible-semaphore/semaphore/util"
 )
 
 type taskPool struct {
-	queue    []*task
-	register chan *task
-	running  *task
+	queue       []*task
+	register    chan *task
+	activeProj  map[int]*task
+	activeNodes map[string]*task
+	running int
 }
 
 var pool = taskPool{
-	queue:    make([]*task, 0),
-	register: make(chan *task),
-	running:  nil,
+	queue:       make([]*task, 0),
+	register:    make(chan *task),
+	activeProj:  make(map[int]*task),
+	activeNodes: make(map[string]*task),
+	running: 0,
 }
 
+type resourceLock struct {
+	lock     bool
+	holder   *task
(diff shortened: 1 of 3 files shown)


matejkramny:
👍 will check this out & test

Thanks!

matejkramny:
Many thanks for this. I'm reviewing if it breaks anything & will be merged shortly after.

My apologies for the long delay.

jfloodnet:
Would be nice to have a 3rd option to allow the same task to deploy concurrently. I've been using semaphore to provision VM's on my proxmox cluster. 

There's no reason I couldn't run the same task concurrently in this scenario and it would have been nice if the option was there.

PR #396, opened 2017-07-12
Feature: Allow managing vault passwords for running playbooks. 

fernandezvara:
This pull request allows to add to the project the required vault passwords to run the playbooks.

Each template allows to select the Vault password to use.

Updated UI for templates and project keys, DB, task runner and `ansible-playbook` invocations.

### DIFF
--- a/db/versionHistory.go
+++ b/db/versionHistory.go
@@ -71,5 +71,6 @@ func init() {
 		{Major: 2, Minor: 3, Patch: 1},
 		{Major: 2, Minor: 3, Patch: 2},
 		{Major: 2, Minor: 4},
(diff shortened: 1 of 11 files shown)


db/versionHistory.go
```diff
@@ -71,5 +71,6 @@ func init() {
 		{Major: 2, Minor: 3, Patch: 1},
 		{Major: 2, Minor: 3, Patch: 2},
 		{Major: 2, Minor: 4},
+		{Major: 2, Minor: 4, Patch: 2},
```
twhiston: Based on semver this is not a patch release as it adds a new feature. It will be in 2.5.0 so please update your migrations to this value here and in the migrations file

PR #433, opened 2017-10-24
feat(template): hidden templates

fiftin:
Add ability to hide templates in the list.

This commit:
- adds button "hide" to each row in the template list. 
- adds button "show hidden" on top of the template list to show hidden templates.

IDs of hidden templates stores to localStorage item "hidden-templates".

### DIFF
--- a/public/html/projects/templates/list.pug
+++ b/public/html/projects/templates/list.pug
@@ -1,7 +1,7 @@
 h3 Task Templates
 	button.btn.btn-success.btn-xs.pull-right(ng-click="add()" style="margin-left: 5px;") New Template
-	button.btn.btn-default.btn-xs.pull-right(ng-if="allShown" ng-click="hideHidden()") Hide Hidden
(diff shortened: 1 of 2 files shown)


fiftin:
Probably not a good idea to use local storage...

twhiston:
What is the use case for this? 
@strangeman - are you using this?

fiftin:
We use a lot of templates in our team. But each team member uses only a few templates.
Yes, we use this feature.

twhiston:
Thanks for the context. If it is in active use then i'm happy to merge it in with the next release. Thanks for the contribution.

fiftin:
I sure this feature is useful. But I don't sure that my implementation is the best. Perhaps local storage is not best place for storing info about hidden templates.

twhiston:
To me this seems like something that should be on the client side, as the application should aim to separate the api and web interface concerns as much as possible. Therefore I have no real problem with it being in the client side storage

fiftin:
@twhiston You convinced me :-)

PR #434, opened 2017-10-25
Use monospaced font & tab indent for JSON editors

fiftin:


strangeman:
Hi! You forgot about the second step in task run dialog:
2017-10-25-164953_1366x768_scrot

fiftin:
Ok, also I've added ng-allow-tab directive to indent in textarea

twhiston:
I think perhaps a better end approach to these json dialogs is to use something like ace code editor library so we can show line numbers and syntax errors in realtime?

fiftin:
I think ACE too complex for simple JSON input. We can validate JSON when user press Save/Run button.

twhiston:
It's incredibly easy to integrate though and the number of issues i've seen on here about how to use the fields that take json suggest to me that we need some extra work in this area that goes beyond the changes here. I'd also in future like to be able to support yml for these sort of things, and having a switchable syntax high lighting would really enhance that. If we validate at save we have to make some changes to support displaying the messages etc.... to me that is more work, and less clear than realtime highlighting would be.
Very happy to discuss this though!

fiftin:
OK, I'll do it

fiftin:
Use ACE instead of it

PR #435, opened 2017-10-25
fix: Cannot read property 'then' of undefined in templates.update

fiftin:


twhiston:
@fiftin What is the actual problem it fixes? (from the vague name it's not something that I have encountered). Could you please provide some context of when this happens

@strangeman - have you used this patch at all?

fiftin:
Each time when ```templates.update method``` called. 
```modal.result.then(callback)``` doesn't return instance of ```modal```, it return result of ```callback``` function.

twhiston:
absolutely! Thanks very much for the explanation (i'm not an FE developer and very much appreciate your contributions in this area). I'll get this merged in soon

twhiston:
@fiftin after merging your other pr this now has a conflict, would be awesome if you could fix it and I will merge this straight in

fiftin:
I fix this bug in next pull request

PR #436, opened 2017-10-26
Add "file" type for inventory

fiftin:


strangeman:
Is this necessary? In my opinion, solutions, that simply duplicates existing `ansible-playbook` options is not good, they make a huge mess with `Extra CLI options`.
We always can pass an inventory file via `Extra CLI options`, no need to make the separate field on a form. This is the way to hell because `ansible-playbook` has around 30 various flags and we cannot add them all, it will make form completely unusable.

fiftin:
I this this is most important flag. I use it each time when I use ansible-playbook. Adding it as ```Extra CLI options``` absolutely unusable.

twhiston:
I think the current 'accepted' way of doing this, a blank inventory and a cli command, seems super hacky and unintuitive. What I like about adding a file filed to this is that essentially then you can upload ec2.py (or whatever) in your repo and pass it as the inventory flag, so this effectively makes it possible to cover all dynamic inventory use cases, without any major changes (or am i wrong about this?) I would suggest though that we remove all the other option in the list as they are disabled anyway ayd this

strangeman:
Okay, your arguments are persuasive. :)

twhiston:
@fiftin can you remove the other inventory type options in the be and fe so that we are left with just static and file?

fiftin:
Done

fiftin:
I found security issue in my solution. User can enter something like this ```../../file.txt```. I want to add inventory path validation but not sure how to do it to support Windows.

api/projects/inventory.go
```diff
@@ -123,6 +125,19 @@ func AddInventory(w http.ResponseWriter, r *http.Request) {
 	mulekick.WriteJSON(w, http.StatusCreated, inv)
 }
 
+func isValidInventoryPath(path string) bool {
+	if currentPath, err := filepath.Abs("./"); err != nil {
```
twhiston: maybe you could use os.Getwd() here instead
https://golang.org/src/os/getwd.go

twhiston: also you might need to look at using https://golang.org/pkg/os/#pkg-constants for the filepath separator to make things cross platform compatible if getwd is not suitable

fiftin:
Added ```gopkg.in/check.v1```

twhiston:
I would prefer if you did not use that test package, it is not really necessary to add it to do the checks you are performing here

fiftin:
What test do you mean?

fiftin:
Why not use unit test framework?

twhiston:
I dont honestly think that you need or want unit test frameworks in go in 99% of cases.

In your test case you can just say
```
if !projects.IsValidInventoryPath("inventories/test"){
  t.Fatal(" path with attibute x should be valid")
}
```
That is all you need!

Something so simple should not be replaced by a third party dependency

fiftin:
Where I should place this test?

fiftin:
IDEA/Golang has UI to run unit tests. IMHO, this is reason to use gotest/gocheck/gobench :)

twhiston:
I also use goland to run my unit tests (and even better code coverage :D), but you dont need anything special for this or any external libraries. IMO we should be implementing tests in as native a manner as possible, and that means doing it in the way the language specification describes, putting them in the same package etc...... as per the above link to the testing package.

fiftin:
Done

api/projects/inventory_test.go
```diff
@@ -21,7 +22,7 @@ func TestIsValidInventoryPath(t *testing.T) {
 		t.Fatal(" a path out of the cwd should be invalid")
 	}
 
-	if IsValidInventoryPath("c:\\test\\inventory") {
+	if runtime.GOOS == "windows" && IsValidInventoryPath("c:\\test\\inventory") {
```
twhiston: I'm not sure that this will ever get run, as it's impossible to test in windows on CircleCI :grimacing: 
But it's good to have as a test anyway!

twhiston:
Just tested this and it is working great with files for inventories (including python scripts) so we now have dynamic inventory support thanks to your efforts. Great work!

PR #449, opened 2018-02-05
Fix stdIn hanging issue

twhiston:
I had an issue with Semaphore where trying to connect to a host that is not already in the known hosts list would ask you if you wished to add them to the known_hosts. This would cause semaphore to be unable to start any other task and a restart would be needed.
With this fix I simply attach a blank string as stdin to the playbook run command, which causes the host key verification to fail in this instance

### DIFF
--- a/api/tasks/runner.go
+++ b/api/tasks/runner.go
@@ -372,6 +372,7 @@ func (t *task) runPlaybook() error {
 	cmd.Env = t.envVars(util.Config.TmpPath, cmd.Dir, nil)
 
 	t.logCmd(cmd)
+	cmd.Stdin = strings.NewReader("")
 	return cmd.Run()
 }
 


twhiston:
@strangeman any thoughts on this?

strangeman:
Interesting. I never had this issue, because we use `ssh_args = -o StrictHostKeyChecking=no` in our `ansible.cfg`.
Seems like a simple and effective solution.

twhiston:
Thanks for the approval :)

Due to our security policy we cant set `StrictHostKeyChecking=no` on our Sempahore machine, so this really bites us if asking for a key hangs the whole program. I've been trying to think if just passing an empty string could ever cause any unintended side effects, but i've never experienced any so I think we can call this ready.

PR #453, opened 2018-02-08
docker/testing/debug refactoring

twhiston:
As promised here is a start in refactoring the project a bit to make it easier to install/test/develop.
There are quite a few changes here but the key points are as follows:

- moved all docker related things to the ./images folder
- used dep to vendor the project
- used a package.json file for node dependencies
- These 2 items together mean that the full install can be totally scripted
- added a testing/development image which builds a fully working environment to develop in and runs 'watch' as standard (see the contributing docs for details)
- made significant changes to the make.sh script to support easier building
- added a dev make.sh script to make some of these actions easier/less verbose

It's also worth noting that we could use all this stuff to increase the CI build speed, and maybe even go for a full docker based test solution (i think this is possible on circle though I don't really use it. I know it is very well supported on codeship though). I'm pretty sure my changes will break the CI as it currently runs, but it should be an easy fix to make.

I'm sure there are still plenty of improvements to make and I would like to take things further by:

- using makefiles/taskfiles instead of bash scripts,
- updating the asset bundling tool
- use gometalinter on the ci
- shellcheck bash scripts
- hadolint dockerfiles
- run tests via goverage and pass coverage to codacy

But this is at least a start. It's important to note that I didn't really make any changes to the code itself, barring the config init function, which is now called explicitly. This is needed as the init can block tests starting without flags set, config files etc....

I'm putting this out there now to get some feedback and what you think about going in this direction, it would also be great if some smart frontend guy could make some comments about if we can manage the public vendor folder with a dependency manager.

Please let me know your thoughts!


strangeman:
Ooh, I strongly agree with using `dep`!

twhiston:
This is a work in progress PR for building sempahore as a deb/rpm for distribution. It uses the awesome https://goreleaser.com project to handle the build and release, this could also be used to do the entire github release setup (including the push to github), so we could remove the release switch from make.sh totally if this agreed upon.

I still need to deal with parsing the changelog nicely and the other finer points of this sort of release but I wanted to put this out there to see what people think and what else you need to make this fully realised. So what else would people like to see in a packaged version of semaphore. Systemd files?

twhiston:
closing this pr as i will introduce these changes in smaller steps, starting with #477

PR #454, opened 2018-02-09
Openshift Template

twhiston:
Just a quick PR to add a template to start Semaphore in openshift (using the openshift mysql-persistent template for storage).
It works when I test it but it would be great if someone else could test it on their own cluster, or in the redhat public cluster. Check the readme in the openshift folder for usage details.

It could also probably do with an nginx container in front of it, could be pretty simple as all the ssl can be done on the edge by openshift, but for now would just be great if some other openshift users could test it.

In future perhaps it would be good to also have a template that builds the image on openshift directly from the repo code.


twhiston:
closing in favor of newer PR #517

Issue #222, opened 2016-11-15
What's the 'Environment' and how to use it?


GoodHaHa:
What's the 'Environment' and how to use it? Thinks a lot.

matejkramny:
It's used to declare environment variables for the playbook, similar to what you would do on the command line

mikequentel:
@matejkramny please, would you show an example? I am trying to set the PATH to include the location of the maven binary, like this: `{"PATH":"$PATH:/opt/apache-maven-3.1.1/bin"}`

matejkramny:
I think it's in the following format, without extra environment key:

```
{
    "PATH": "{{ansible_env.PATH}}:/opt/apache-maven-3.1.1/bin"
}
```

mikequentel:
Hi @matejkramny many thanks for the tip...the values are not getting exported to underlying shell scripts, but the syntax correctly parses. Now I am trying this: `{"PATH": "{{ansible_env.PATH}}:/opt/apache-maven-3.1.1/bin","MAVEN_HOME": "/opt/apache-maven-3.1.1/bin"}` -- Semaphore appears to successfully consume that, but shell scripts called from the playbook do not recognise the `MAVEN_HOME`, for example. But that is probably due to the shell scripts being launched in another process. I shall keep hacking at this.

Many thanks for your help and advice. Semaphore is a great tool.

matejkramny:
I'm happy it works for you. Let us know if you need anything

PR #458, opened 2018-02-14
Add environment key ENV 

twhiston:
This adds the optional key ENV to the environment that allows you to pass command execution level arguments to the run.
This is useful if you need to alter something such as the ansible.cfg path, which needs to be done via env var due to the working directory for semaphore being set to the playbook/tmp path.
This would also satisfy the final request of the (closed issue) #222 

I'm not sure if this is the way that people want to go with this sort of feature, but having a special key in the environment json seemed like the lowest cost way to achieve this without db migrations and ui updates etc.... If people feel it muddies the waters regarding the purpose of the environment file then I will happily implement it in a different way


twhiston:
@strangeman any thoughts about this? I think we should consider adding a way to pass environmental variables to the command execution environment, rather than just variables to ansible, but this is probably not the way to do it! Should we add another field to the task/task run dialogs that allow you to specify environment variables? Any better ideas?

strangeman:
Uh, this is the strange feature. I worked more than 3 years with Ansible, in different environments and never used external ENV variables for it. In my personal opinion, this is a rarely used feature, so we may implement it in simple (and a bit dirty) way.

twhiston:
without this feature how would you pass a custom ansible.cfg file? putting them in the project root is not a good enough solution as you may have multiple playbooks in a single repo, subfolders etc.... this is the specific use case i did this for as it's something we needed at work. I would be very open to a better solution though 😀

strangeman:
Hmm... we just store it in the project root. :) We have multiple playbooks, but `ansible.cfg` contains only basic information. Some variables (like remote_user) overridden via variables, another - via flags.

twhiston:
we have a specific use case then, as we store all our openshift cluster version playbooks in the same repo and have to use different ansible.cfg files to pass distinct plugin folder locations for each.
So would you support doing it this way for now (and an addition to the docs clearly explaining it, maybe also a ui addition too?) or shall i think about a more robust implementation?

strangeman:
Let's implement it as is. If the users will request the more clean way to pass ENV variables - we'll make another task.

twhiston:
ok, i've additionally fixed the issues from codacy and added to the environment interface text to mention the ENV key. I will probably also add something in the wiki in future but rewriting that is a separate task I think.

twhiston:
@strangeman, can you give this a quick once over and if you are happy I will merge it

Issue #460, opened 2018-02-15
Semaphore should create the ephemeral dir if it does not exist


twhiston:
If the semaphore tmp/playbook dir does not exist or semaphore does not have access to it playbooks will refuse to run as the inventory, keys etc... cannot be written. At least the creation should be checked at runtime, as if a tmp dir is specified a machine reboot will legitimately destroy it.
Do we want to check/fix ownership as well? or should this be an error?

twhiston:
permissions are a tricky thing here, as linux wants open permissions with a sticky bit for the /tmp folder so we have to make that the default. If you are not writing to a temp folder you could always control this via some other automation on the runtime server, or make systemd create it in the ExecStartPre.

PR #461, opened 2018-02-15
Create tmp dir if not existent before job run

twhiston:
This PR is for #460 it adds a function to check if the tmpDir exists at the start of a task run, and if not attempts to create it. If the creation fails it will return an error and halt execution of the task


api/tasks/runner_test.go
```diff
@@ -0,0 +1,67 @@
...
+	"math/rand"
+	"time"
+	"os"
+)
+
+
+func TestCheckTmpDir(t *testing.T) {
+
+	//It should be able to create a random dir in /tmp
+	dirName := os.TempDir()+ "/" + randString(rand.Intn(10 - 4) + 4)
```
strangeman: Why you're don't use the simple https://golang.org/pkg/io/ioutil/#TempDir? As I see, it creates the random name (after prefix) too.

api/tasks/runner.go
```diff
@@ -439,3 +446,15 @@ func (t *task) envVars(home string, pwd string, gitSSHCommand *string) []string
...
 	return env
 }
+
+// checkTmpDir checks to see if the temporary directory exists
+// and if it does not attempts to create it
+func checkTmpDir(path string) error {
+	var err error = nil
+	if _, err := os.Stat(path); err != nil {
+		if os.IsNotExist(err) {
+			return os.MkdirAll(path, 1777)
```
strangeman: Is it safe? If we want to store keys in the temporary location they may be found with simple `find`, if we make this directory world-readable.

api/tasks/runner_test.go
twhiston: Because I dont want to create the directory at this point. I want to test that my function is able to create a directory when it does not exist

api/tasks/runner.go
twhiston: well... this is an issue actually because if you are writing to tmp it must be 1777.
However it would be simple enough to see if the first part of the path matches the os.TmpDir or "/tmp" (because os.TmpDir could be $TMPDIR) and then change the permissions accordingly. I would suggest either 0660 or 0640 for non tmp folders, what do you think?

twhiston:
So i spoke to some sysadmins and made some tests in our production environment and the best set of permissions I can use for the tmp folder are 0700, which is pretty locked down and should protect any secrets that we write or pull. @strangeman with this permission setting would you be happy with approving this PR?

strangeman:
Yeah, `0700` is great.

Issue #379, opened 2017-06-19
Use the UI to prompt for user and sudo/ssh password


joernott:
For security reasons, I am not allowed to use passwordless sudo and the ssh connect to the remote machines should ideally happen with my user, not a shared "semaphore" user.

I have seen that there is an option "Extra CLI Arguments", but I would like to use "-K" (--ask-become-pass) and -u to specify the remote user and the password for sudo (ideally using the user name which is using the UI). Is there a way to achieve this kind of behaviour?

matejkramny:
Not currently, however there are ideas in #85, #115, #113..

joernott:
I have been thinking about how all these can be solved in a consistent way. I believe,a valid approach would be:
Making use of variables solves a lot of problems (e.g. setting usernames, passwords). Host specific variables can already be set in inventories. There should be a way to pre-define variables from the context (e.g. username, password from the currently logged in user (useful for ask_become_pass when using LDAP). We should have a way of setting variables (or telling semaphore to prompt for them) in task templates. We also should have a way of specifying a users ssh key (including a prompt for the keys password) and a user specific vault. Of course, it is the responsibility of the user to actually make sure, that the vault contains the information needed.

Thoughts:
- Passing ansible variables via "--extra-vars="@extravars_file" is preferable over specifying them at the commandline and thus exposing them to the process list. There is no extra-var for the vault password though, so we have to use the vault password file.
- Vault reduces the number of passwords to be sent to one, the rest can be set from the user specific variables
- We either have to remember passwords when logging in or have to provide them on every run of a task
- Vault password files should be deleted after running the task. Ideally they should be stored in a folder which can be mounted as tmpfs so that no traces on disk remain

I have forked the project and I will try to implement the above. At the moment, it looks like a rather large feature. Lets see what happens
RFC

matejkramny:
@joernott semaphore already utilises `--extra-vars` and passes the json environment.

You're describing a number of features:

- Passing additional envirnoment variables to ansible
- Semi-interactive terminal ?
- Vault enhancements

Matej

joernott:
Hello Matej,

I have seen, that you already use extra-vars. But currently, these are defined in a template and are open to all users.

[...]

PR #467, opened 2018-02-24
WIP Issue #379 user specific vars and user specific vault

joernott:
As a followup to https://github.com/ansible-semaphore/semaphore/issues/379, I started to add the features to have user specific variables and vaults.


twhiston:
Hi @joernott, thanks for contributing this, seems like there is some issue with migrations currently. I was hoping to merge in #396 this week once I start to add all the past PR's so it would be great if you could ensure the changes related to vault play nicely with that PR, or if that PR will cause issues for you please let me know so we can find the right path forwards with this feature.

As this is quite a large change i'm going to mark it with 2.6.0 so we have the time to make sure it is right, but of course if it gets finished sooner we will get it in to 2.5.0.

joernott:
I have merged #396 and #379 in https://github.com/ansible-semaphore/semaphore/pull/473 which is based on the current develop branch. All merge conflicts should be resolved now and the requested changes to the version numbers have been done.

PR #470, opened 2018-02-28
Add binaries to gitignore and remove binddata.go

fiftin:
I found ```util/binddata.go``` in ```.gitignore``` but file isnt't deleted. I've deleted it.

Also added ```cli/semahore_*``` to gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ node_modules
 
 .idea/
 caddyfile
+cli/semaphore_*


twhiston:
You shouldn't really be putting your semaphore_config.json in the cli folder anyway (use make.sh which expects the file in the repo root, or pass the config file option when starting it), so the ignore seems un-necessary. Please remove this.

However good catch on the bindata, looking through the commit history it seems this has been in the repo for a while, when it should actually only get generated by the make.sh during release and other processes.

Issue #463, opened 2018-02-16
Api endpoints extra parameters


twhiston:
arising from #190 it would be good if we could POST the same options to the api to start a task that are available from the interface (environment etc....)

kumy:
We are also interested in passing dynamically `Extra CLI arguments` to the `post /project/{project_id}/tasks ` endpoint.

kumy:
As a workaround I've used the `put /project/{project_id}/templates/{template_id}` endpoint to update the `Extra CLI arguments` before launching the task…

lennart:
I had been using the api to start a job in v2.6.8 with exactly the mentioned functionality:

with new_task.json being:

```json5
{
  "template_id": 1,
  "debug": false,
  "dry_run": false,
  "arguments": "[\"--extra-vars\", \"foo_arg=bar baz_arg=qux\"]"
}
```

then ran:

```
cat << EOF > semaphore_api_headers
Content-Type: application/json
Accept: application/json
Authorization: Bearer ${SEMAPHORE_DEPLOY_TOKEN}
EOF
curl --verbose --fail --show-error --header @semaphore_api_headers -d@new_task.json ${SEMAPHORE_PROJECT_URL}/tasks
```

and this functionality (`"arguments": "[\"--extra-vars\" ...]"`) is now missing in 2.8.9 and I could not find any note of breaking (api) changes.
What was the reason this was removed and more importantly, how can we get it back?

I'll read the code and try to figure out how to restore this functionality (I'd call this regression)

I would agree that the approach described above is somewhat clunky and it might be nicer to specify extra-args as JSON directly.

lennart:
ok, as far as I understand, this now goes into the `environment` key, see: https://github.com/ansible-semaphore/semaphore/issues/767

lennart:
however, with the former usage via `arguments` I could still use a defined environment in the task template and have the `arguments` be merged with this. the _new_ version completely overrides any configure environment of the task template, i.e. I have to pass in all variables on task invocation, whereas before I only had to pass in those that need to be overriden/vary between job runs.

Is there a way to achieve the _merging_ of environments/extra-vars like before?

fiftin:
Hi @lennart ,

Yes, you are right, environments must be merged. It will be implemented ASAP.

lennart:
Hi @fiftin,

great, thanks!

fiftin:
All done.

Issue #462, opened 2018-02-16
Reduce when panic is called


twhiston:
Currently semaphore panics a lot, which crashes the whole app. This is especially problematic in docker environments. IMO even if the db is not available (a common panic cause currently) the application itself should not crash. If within the context of a task it should halt the job, but the application itself should not exit.

strangeman:
```
panic: Post https://api.telegram.org/botXXXXXXXX:XXXXXXXXXXXXXXXXXXX/sendMessage: dial tcp: lookup api.telegram.org on 104.238.186.189:53: read udp 192.168.0.71:47609->104.238.186.189:53: i/o timeout

goroutine 387 [running]:
github.com/ansible-semaphore/semaphore/api/tasks.(*task).sendTelegramAlert(0xc420b68000)
	/mnt/src/sources/go/src/github.com/ansible-semaphore/semaphore/api/tasks/alert.go:101 +0x448
github.com/ansible-semaphore/semaphore/api/tasks.(*task).fail(0xc420b68000)
	/mnt/src/sources/go/src/github.com/ansible-semaphore/semaphore/api/tasks/runner.go:40 +0x68
github.com/ansible-semaphore/semaphore/api/tasks.(*task).prepareRun(0xc420b68000)
	/mnt/src/sources/go/src/github.com/ansible-semaphore/semaphore/api/tasks/runner.go:111 +0x618
created by github.com/ansible-semaphore/semaphore/api/tasks.(*taskPool).run
	/mnt/src/sources/go/src/github.com/ansible-semaphore/semaphore/api/tasks/pool.go:77 +0x4e1
```

Network connection troubles should be handled w/o panic and crash.

matejkramny:
hmm agreed. Maybe we could keep the panics and implement a recover at the top of the routing stack.

Re database connection at the start, are we proposing to let it sit and wait until the db can be connected?

fiftin:
Done

PR #473, opened 2018-02-28
Feature/both vaults

joernott:
Unifying PR#379 and PR#396 and resolve merge conflicts and changes as requested


api/users.go
```diff
@@ -24,9 +24,12 @@ func getUsers(w http.ResponseWriter, r *http.Request) {
 
 func addUser(w http.ResponseWriter, r *http.Request) {
 	var user db.User
+
 	if err := mulekick.Bind(w, r, &user); err != nil {
 		return
 	}
+<<<<<<< HEAD
```
twhiston: unresolved merge conflict has been left in here

joernott: Saw it, too. I just wasn't fast enough to push before you found it...

twhiston: 😁 just happened to be around as you pushed, thanks for the fix. I'll try to work out why circle is failing at the moment

twhiston:
It will probably take me a few days to review this properly, i'll do my best to get it done over the weekend. Thanks for your effort

db/versionHistory.go
```diff
@@ -72,5 +72,6 @@ func init() {
 		{Major: 2, Minor: 3, Patch: 2},
 		{Major: 2, Minor: 4},
 		{Major: 2, Minor: 5},
+		{Major: 2, Minor: 6},
```
twhiston: please merge this with the existing 2.5 migration as they will come out in the same release

util/cache.go
```diff
@@ -0,0 +1,111 @@
...
+package util
+
+import (
+	"errors"
+	"time"
+
+	"github.com/patrickmn/go-cache"
+)
+
+type CachedUser struct {
```
twhiston: Please comment all the exported items in this class, as per the go styling standard (gometalinter can help with finding these sort of style issues) https://blog.golang.org/godoc-documenting-go-code

util/cache.go
```diff
@@ -0,0 +1,111 @@
+// Handle the password caching
```
twhiston: package comment should be of the form "Package util ..."

util/util_test.go
```diff
@@ -29,3 +29,143 @@ func mockParam(w http.ResponseWriter, r *http.Request) {
 
 	w.WriteHeader(200)
 }
+
+func TestUserCache(t *testing.T) {
```
twhiston: please put this test in to cache_test.go instead (and shout out for actually adding tests! :1st_place_medal: )

util/cache.go
```diff
@@ -0,0 +1,111 @@
...
+package util
+
+import (
+	"errors"
+	"time"
+
+	"github.com/patrickmn/go-cache"
```
twhiston: Is there a reason that you chose this key:value store in particular? It has a lot of open issues and pr's and it doesnt have a lot of current activity, so I dont feel super comfortable in adding it as a dependency to the project. Why not use something well known like bolt (https://github.com/coreos/bbolt) or another? There is a good list of these here https://github.com/avelino/awesome-go#resource-embedding

api/login.go
```diff
@@ -194,7 +195,7 @@ func login(w http.ResponseWriter, r *http.Request) {
 		Value: encoded,
 		Path:  "/",
 	})
-
+	util.UserVaultCache.AddEntry(login.Auth, login.Password)
```
twhiston: you are not checking for errors returned here

api/tasks/runner.go
```diff
@@ -172,6 +199,9 @@ func (t *task) run() {
 		t.fail()
 		return
 	}
+	var user db.User
+	t.fetch("User not found", &user, "select * from user where id=?", t.task.UserID)
+	util.UserVaultCache.DecrementUsage(user.Username)
```
twhiston: no error checking

twhiston: if a failure does not need to stop program execution then please at least log it.

api/tasks/vault.go
```diff
@@ -0,0 +1,50 @@
...
+	"strconv"
+
+	"github.com/ansible-semaphore/semaphore/db"
+	"github.com/ansible-semaphore/semaphore/util"
+)
+
+func (t *task) InstallVaultFiles() error {
+	var user db.User
+
+	if t.template.UserVault == true {
```
twhiston: you can simplify this to `if t.template.UserVault {`

api/tasks/vault.go
```diff
@@ -0,0 +1,50 @@
...
+		if err != nil {
+			return err
+		}
+		t.log("Installed vault file " + util.Config.TmpPath + "/vault_" + strconv.Itoa(t.task.ID))
+	}
+	return nil
+}
+
+func (t *task) RemoveVaultFiles() error {
+	if t.template.UserVault == true {
```
twhiston: you can also simplify this

api/projects/templates.go
```diff
@@ -94,16 +100,21 @@ func AddTemplate(w http.ResponseWriter, r *http.Request) {
 
 	var template db.Template
 	if err := mulekick.Bind(w, r, &template); err != nil {
+		fmt.Println(err)
 		return
 	}
-
-	res, err := db.Mysql.Exec("insert into project__template set ssh_key_id=?, project_id=?, inventory_id=?, repository_id=?, environment_id=?, alias=?, playbook=?, arguments=?, override_args=?", template.SshKeyID, project.ID, template.InventoryID, template.RepositoryID, template.EnvironmentID, template.Alias, template.Playbook, template.Arguments, template.OverrideArguments)
+	if template.UserKey == true {
```
twhiston: simplify please

api/projects/templates.go
```diff
@@ -128,14 +140,20 @@ func UpdateTemplate(w http.ResponseWriter, r *http.Request) {
...
+		fmt.Println(err)
 		return
 	}
 
 	if template.Arguments != nil && *template.Arguments == "" {
 		template.Arguments = nil
 	}
 
-	if _, err := db.Mysql.Exec("update project__template set ssh_key_id=?, inventory_id=?, repository_id=?, environment_id=?, alias=?, playbook=?, arguments=?, override_args=? where id=?", template.SshKeyID, template.InventoryID, template.RepositoryID, template.EnvironmentID, template.Alias, template.Playbook, template.Arguments, template.OverrideArguments, oldTemplate.ID); err != nil {
+	if template.UserKey == true {
```
twhiston: simplify please

api/tasks/http.go
```diff
@@ -26,10 +26,14 @@ func AddTask(w http.ResponseWriter, r *http.Request) {
 	taskObj.Status = "waiting"
 	taskObj.UserID = &user.ID
 
-	if err := db.Mysql.Insert(&taskObj); err != nil {
+	if err := util.UserVaultCache.IncrementUsage(user.Username); err != nil {
 		panic(err)
 	}
 
+	if err := db.Mysql.Insert(&taskObj); err != nil {
+		util.UserVaultCache.DecrementUsage(user.Username)
```
twhiston: no error checking

api/tasks/http.go
```diff
@@ -26,10 +26,14 @@ func AddTask(w http.ResponseWriter, r *http.Request) {
...
+		panic(err)
```
twhiston: do we REALLY need to panic here? Semaphore is way too panicky at the moment, and that's very problematic when running in docker. It's on the cards to make it generally less panicky (#462) so i'd prefer not to introduce more. Can we soft fail in some nice way here and just log the error, or at least recover from the panic?

api/tasks/http.go
```diff
@@ -43,6 +47,7 @@ func AddTask(w http.ResponseWriter, r *http.Request) {
 		ObjectID:    &taskObj.ID,
 		Description: &desc,
 	}.Insert()); err != nil {
+		util.UserVaultCache.DecrementUsage(user.Username)
```
twhiston: no error checking

api/tasks/runner.go
```diff
@@ -33,21 +34,38 @@ type task struct {
...
 
 func (t *task) fail() {
+	var user db.User
 	t.task.Status = "error"
 	t.updateStatus()
 	t.sendMailAlert()
 	t.sendTelegramAlert()
+
+	t.fetch("User not found", &user, "select * from user where id=?", t.task.UserID)
+	util.UserVaultCache.DecrementUsage(user.Username)
```
twhiston: no error checking on any of this and the 2 lines after

api/tasks/runner.go
```diff
@@ -225,15 +255,32 @@ func (t *task) populateDetails() error {
 	}
 
 	// get access key
-	if err := t.fetch("Template Access Key not found!", &t.sshKey, "select * from access_key where id=?", t.template.SshKeyID); err != nil {
-		return err
+	if t.template.UserKey == true {
```
twhiston: can be simplified

api/tasks/user.go
```diff
@@ -0,0 +1,39 @@
...
+			t.log("Error writing user vars file " + util.Config.TmpPath + "/user_vars_" + strconv.Itoa(t.task.ID) + ": " + err.Error())
+			return err
+		}
+		t.log("Created user vars file " + util.Config.TmpPath + "/user_vars_" + strconv.Itoa(t.task.ID))
+	}
+	return nil
+}
+
+func (t *task) RemoveUserVarFile() error {
+	if t.template.UserVault == true {
```
twhiston: can be simplified

public/html/projects/templates/add.pug
```diff
@@ -39,6 +45,11 @@
 			.col-sm-6
 				select.form-control(ng-model="tpl.environment_id" ng-options="env.id as env.name disable when env.removed for env in environment")
 					option(value="") -- Select Task Environment --
+		.form-group
+			label.control-label.col-sm-4 Vault Password
```
twhiston: This currently does not seem to filter on the vault password key type properly and shows all keys

public/html/projects/templates/add.pug
```diff
@@ -20,10 +20,16 @@
...
-					option(value="") -- Select SSH Key --
+			label.control-label.col-sm-4 SSH key
+			.col-sm-6.col-sm-offset-4
+				.radio: label
+					input(type="radio" ng-model="tpl.user_key" name="user_key" data-ng-value="false")
+					select.form-control(ng-model="tpl.ssh_key_id" ng-options="key.id as key.name disable when key.removed for key in keys" ng-disabled="tpl.user_key==true")
+						option(value="") -- Select SSH Key --
+				.radio: label
+					input(type="radio" ng-model="tpl.user_key" name="user_key" data-ng-value="true")
+					| User specific SSH Key
```
twhiston: I'm not sure this is the right way to do this, because I can create more than 1 ssh key that belongs to me in a project, as they do not appear in the dropdown above I would expect that they get listed some other way, OR that it would be impossible for me to create more that 1 personal ssh key per project (and this was clearly explained by the dropdown where you pick the user who owns the key)

util/cache.go
```diff
@@ -0,0 +1,111 @@
...
+}
+
+type VaultCache struct {
+	store *cache.Cache
+}
+
+var UserVaultCache *VaultCache
+
+func NewVaultCache() *VaultCache {
+	var vc *VaultCache
```
twhiston: you don't need to declare this var, just do the assignment with := on the next line

util/cache.go
```diff
@@ -0,0 +1,111 @@
...
+		return nil, errors.New("User not cached")
+	}
+	cu = item.(*CachedUser)
+	return cu, nil
+}
+
+func (vc *VaultCache) AddEntry(User string, Password string) error {
+	cu, err := vc.Get(User)
+	if err == nil {
+		if cu.Password != Password || cu.LoggedIn == false {
```
twhiston: `!cu.LoggedIn`
Please check for other places in this file where you do the same

util/cache.go
joernott: Essentially, I don't want to store the decrypted user password anywhere on disk, so I was looking for a non-persistent key-value store which keeps things in memory. go-cache is thread-safe and handles things in memory. I have not yet used bbolt, so I am not sure, if we could use it  as a in-memory-only Store

twhiston: Ok, in-memory makes total sense for that use case. There are a few in memory suggestions here https://github.com/gostor/awesome-go-storage that maybe you could check out to see if they are better maintained.

joernott: OK, from looking at it, we have BuntDB, which is also unmaintained, bigcache, which seems a bit oversized, https://github.com/bluele/gcache and  https://github.com/muesli/cache2go, which look like suitable replacements. I would go with cache2go, if you agree.

api/tasks/vault.go
joernott: yes, that is a bad habit working with languages like PHP, where you need to do things like === to make sure, the boolean you have is actually a boolean because something like "if $somevar" behaves differently from "if $somevar === true". I'll go through the code and fix that.

util/cache.go
twhiston: cache2go looks good to me, thanks :)

api/tasks/http.go
joernott: I didn't care, because we are in the middle of error handling. As we are panicking in the next line anyways, a cleanup does not make sense at all. However, tasks/http.go has a lot pf panic in there which should be replaced by appropriate logging (see comment above). So I added a basic error handling for the change. I'll leave the panic alone as we should handle this in #462

api/tasks/http.go
joernott: The AddTask function does not have any error handling but panicking in every error case. If we can't create a task, we have nothing left to do but to throw an error.
For another application (https://github.com/joernott/go-ampel), I used to reply with the correct http status codes instead of failing. I would recommend to do that in a consistent way in a separate PR for #462 for the whole http.go  instead of just doing it for this one call.

twhiston: Throwing an error and logging it is definitely correct, imo panicking is not (though as previously stated I am aware we do this too often atm), because a panic crashes the entire program, including the web interface, which I do not think should be the case for any error (the be and fe should effectively be totally decoupled even though they are not currently).

I also agree that we should handle this in a consistent way i'm just reluctant to let in more code that we immediately need to change due to panic handling as we are creating technical debt. 
Would you be interested in tackling #462 after this feature is added? That would give me more peace of mind ;)

joernott: I can't promise how fast I am going to solve #462 that because I am working on this in the evenings or when I have to wait for my co-workers in my current project. So I can't really predict the time I cam spend per day/week, but I will definitively give it a go.

As I am running semaphore inside a docker container (one conforming to my own guidelines), I definitely need less panic and more clever error handling. For the moment, docker swarm takes care of just restarting the container and the application doesn't panic that often, so the impact on my tiny production environment is negligable. However, running it in large scale multi-user environment with hundreds of servers and many concurrent tasks requires a very stable semaphore to be tagged as production ready.

db/versionHistory.go
joernott: Can you be more specific as to what other PRs I should merge? At the moment, it looks only like #463, as #457 does not contain any go code. #396 is merged.

api/tasks/http.go
twhiston: ^ Absolutely agreed! At work we run it with systemd so it restarts, but I run it for myself in openshift, which although it takes care of the restarts, means that I see a constant huge count of container restarts, and loose my metrics every time etc..... So it would be really good to solve this. I might also take a crack at  #462 so maybe we can collaborate on that together, and as you are a docker user I might give you a mention when I need docker stuff testing if that would be ok? I'm planning to rework the docker containers in the near future anyway so we can have development builds, are not restricted to pulling bundles from github releases etc.... so I would definitely appreciate someone taking a look at these changes when they happen.

db/versionHistory.go
twhiston: Version should be set to 2.5 here and you should merge your sql migration with the v2.5.0.sql file which is already in the development branch migrations folder.

api/tasks/http.go
joernott: I am happy to help and work together with you on those two.

db/migrations/v2.6.0.sql
```diff
@@ -0,0 +1,22 @@
+ALTER TABLE access_key ADD COLUMN `owner` int(11) DEFAULT 0;
+UPDATE access_key SET owner=0;
+#SET @SAVE_sql_mode = @@sql_mode;
```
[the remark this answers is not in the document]
strangeman: Looks like codacy not properly working with sql.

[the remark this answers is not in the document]
twhiston: Yeah. I will tell it not to analyse these files due to all the false positives. Sorry for the noise created while we get it configured correctly

megakoresh:
Is this PR abandoned? I am evaluating using this semaphore in our company and this is one of the features required.

megakoresh:
Its July already. Is this PR abandoned? I am evaluating using this semaphore in our company and this is one of the features required. @joernott are you still working on this? Also shouldn't https://github.com/ansible-semaphore/semaphore/pull/467 and https://github.com/ansible-semaphore/semaphore/pull/396 be closed in favour of this?

gaby:
@twhiston @fiftin @joernott Any updates regarding this PR?

fiftin:
@gaby No updates. But I planning to work on it in the near future.

ziptx:
Apologies for not knowing git process and reporting an error.   I see this feature in the code of the v2.7.16 release.   The "New Template" in the GUI has the "Vault Password" dropdown.   The GUI is not currently saving the "vault\_pass\_id" into the "project\_\_template" table.   \[Fails on both a creating a new template and updating a template\].   If the correct "id" from the "access\_key" table is manually added to the "vault\_pass\_id" in the "project\_\_template" record, the vault password functionality works as expected when running the template/playbook.    The insert/update code is correct in templates.go on lines 101 & 139.  I would guess the "vault\_id" /template.VaultID variable is not getting passed into the module.

PR #477, opened 2018-03-07
Change build process to use task

twhiston:
This is a w.i.p to change from the make.sh build format to using task (https://github.com/go-task/task)
Made as a w.i.p so I can work on the ci part of the change and so that you can get a feeling for if you like this direction or not.

I prefer task to a Makefile (or a bash script as we have now) as it simplifies things a lot, it's a simple declarative yaml style, it allows for caching and dependencies (though I will configure these later), and enhances the readability of the make process, it also uses golang templating in the commands, which can make it very powerful.

Probably the bit that does not currently work is the release section, but this is also something I would like to propose a change to after this PR is dealt with (moving to use go-releaser https://github.com/goreleaser/goreleaser ) so i'm not overly worried about getting that part correct at this time


Taskfile.yml
```diff
@@ -0,0 +1,122 @@
+deps:
+  desc: Install all dependencies
+  cmds:
+   - task: deps:be
+   - task: deps:fe
+   - task: deps:watch
+
+deps:be:
```
strangeman: `be/fe` = `backend/frontend`? Never seen this abbreviation before. :)

twhiston: yeah exactly, we always use these abbreviations, but if you feel it is unclear i'm happy to change it

twhiston:
removed the w.i.p label as you seem happy with the approach. I will add some extra comments and change the contribution guide to reflect the changes and then ask again for approval.

twhiston:
@strangeman I've updated the docs and squashed this to a single commit, so i think it should be good to go. Let me know if you want any changes

twhiston:
Ignore the label, this is not merged, i did a force push and github went a bit crazy!

Taskfile.yml
strangeman: nop, I'm ok with it :)

PR #478, opened 2018-03-08
Taskfile

twhiston:
@strangeman sorry this ended up in a new PR, github would not let me change the status of the PR that was not merged (#477) so I couldnt put it back somewhere that was visible for you to review :roll_eyes:


twhiston:
I've updated this to the new version 2 taskfile format as that was released a couple of days ago before i merged it, so make sure to go get task again before you use it!

PR #479, opened 2018-03-09
Update angularjs/angular-ui to last version

fiftin:
Update angularjs to v1.6, angular-ui to 2.5.

I replaced ```success/error``` blocks to ```then/catch``` because angular v1.6 isn't support it.

Please apply ```update-angular``` PR to ```semaphore-vendor``` before

### DIFF
--- a/public/html/projects/templates/add.pug
+++ b/public/html/projects/templates/add.pug
@@ -52,7 +52,6 @@
 				.checkbox(uib-tooltip="Usually semaphore prepends arguments like `--private-key=/location/id_rsa` to make sure everything goes smoothly. This option is for special needs, where semaphore conflicts with one of your arguments."): label
 					input(type="checkbox" ng-model="user.admin")
 					| Override semaphore arguments
-				p.help-block
 .modal-footer
 	button.btn.btn-default.pull-left(ng-click="$dismiss()") Dismiss
 	button.btn.btn-danger(ng-if="tpl.id" ng-click="$close({ remove: true })") remove
--- a/public/html/projects/templates/add.pug
+++ b/public/html/projects/templates/add.pug
@@ -44,15 +44,15 @@
 
 		hr
 		.form-group
-			label.control-label.col-sm-4 Extra CLI Arguments (*MUST* be a JSON array!)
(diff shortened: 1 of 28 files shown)


twhiston:
:fire: i'll test this over the weekend and get back to you, nice one!

Unrelated frontend question -  is it possible to create a folder structure where the assets that we compile from all live in a different root to the output? currently we mix in the pug and html files, and the less and css, during the generation process. 
Would it be possible to refactor in such a way that we separate all the files actually needed to run the app from the files needed to build it? I guess the bundled js needs to stay where it is (or maybe i am wrong, my frontend skills are not strong :stuck_out_tongue: ) but it would be cool if we could create a bit more granularity here if possible. What do you think?

fiftin:
Yes, we can move ```pug/less``` files from ```semaphore/public/html``` & ```semaphore/public/css``` to directories ```semaphore/less``` & ```semaphore/pug``` for example.

fiftin:
or to ```semaphore/public/pug``` & ```semaphore/public/less```

twhiston:
well i would like to keep the frontend completely seperate from the backend, because really it is kind of a separate project from the API (conceptually at least).
So maybe we could have another subfolder something like

```
root
- api
- cli
- etc.....
- web
    - resources
       - pug
       - less
       - unbundled js etc....
    - public
       - html
       - css
       - js
```
That way we can just use packr on the public folder and put only the things we need in to the binary. How does that feel to you. @strangeman any thoughts?

strangeman:
Good idea. Current folder structure a bit annoying, I often get confused with pug and html files.

fiftin:
I ready to do this but in next PR. OK?

twhiston:
Yeah would be great in another PR, thanks! I'll give this one a review tomorrow so we can get it merged

fiftin:
Codeatyle issues above are not relevant to this PR.
Can we fix them after PR application?

twhiston:
@fiftin yeah absolutely fine to fix codacy issues after this PR, i just enabled it so it reviewed everything as soon as I did!

twhiston:
looking good to me, merging

PR #482, opened 2018-03-10
Version generator

twhiston:
This is a proposal to replace the version.go file with a generated version, which is created at `task compile` time. This allows us to give every release of semaphore a unique ID by which we can identify it, which becomes more important once we start pushing out docker images for development builds, and it also makes it easier to identify what build you are actually running. It additionally adds the `--version` option to the semaphore cli, which will return the current version string and exit.

The version generator itself is really simple, it expects you to pass in a version string to it, and renders the file pretty much as before, therefore most of the actual logic happens in the taskfile to render a unique value. It uses the following logic:

If a tag exists on this commit then use it.
Else if the `SEMAPHORE_VERSION` variable is set use it
Else build an id using the following:
`{{ .BRANCH }}-{{ .SHA }}-{{ .TIMESTAMP }}{{ if .DIRTY }}-dirty{{ end }}`


twhiston:
closed this so i can rebase and clean up commit history

PR #489, opened 2018-03-12
Change Frontend directory structure

fiftin:
* Split ```public``` to ```web/public``` & ```web/resources```
* Replace ```ansible-vendor``` to npm deps
* Fix build under Windows


Taskfile.yml
```diff
@@ -51,25 +49,29 @@ tasks:
...
-      - pug $(find ./html/ -name "*.pug")
-      - node ./bundler.js
+      - public/css/semaphore.css
+      - public/html/*.html
+      - public/html/**/*.html
+      - bundle.json
+    cmds:
+      - mkdir -p public/css
+      - mkdir -p public/html
+      - mkdir -p public/fonts
```
twhiston: can we keep these directories in the repo with a .gitkeep so we don't need to try and make them every time (as they will usually exist)

Taskfile.yml
```diff
@@ -51,25 +49,29 @@ tasks:
...
+      - lessc resources/less/semaphore.less > public/css/semaphore.css
+      - pug resources/pug --out public/html
+      - cp ../node_modules/font-awesome/fonts/* public/fonts
+      - node bundler.js
 
   compile:be:
     desc: Runs Packr for static assets
     sources:
-      - public/*
+      - web/resources/*
```
twhiston: this isnt going to work because packr wants to pack the final set of assets, ie the public folder and not what we build it from

Taskfile.yml
```diff
@@ -51,25 +49,29 @@ tasks:
...
+      - public/html/*.html
+      - public/html/**/*.html
+      - bundle.json
+    cmds:
+      - mkdir -p public/css
+      - mkdir -p public/html
+      - mkdir -p public/fonts
+      - lessc resources/less/semaphore.less > public/css/semaphore.css
+      - pug resources/pug --out public/html
+      - cp ../node_modules/font-awesome/fonts/* public/fonts
```
twhiston: can we keep node_modules inside the 'web' folder?

Taskfile.yml
```diff
@@ -30,15 +30,13 @@ tasks:
 
   deps:fe:
     desc: Installs npm requirements for front end from package.json
-    dir: public
+    dir: web/resources
     cmds:
-      - git submodule update --init --recursive
       - npm i -g less pug-cli
```
twhiston: is it possible to add these tools to the devDependencies of the package.json so we don't need to install them here?

api/tasks/runner_test.go
```diff
@@ -23,6 +23,11 @@ func TestCheckTmpDir(t *testing.T) {
 		t.Fatal(err)
 	}
 
+	if runtime.GOOS == "windows" {
+		// Windows doesn't support os.Chmod, need use acl.Chmod
+		return
```
twhiston: rather than just skipping the tests on windows can you rewrite them to work, or remove this change and make a new issue about it so we can work on it separately?

Thanks!

Taskfile.yml
fiftin: This global packages, we can't put these to package.json

api/tasks/runner_test.go
fiftin: I think this test is incorrect. It tests limitations of the file system. Maybe remove it?

fiftin: This test will fail under Linux if Temp directory located on NTFS partition for example.

Taskfile.yml
fiftin: ok, moved ```package.json``` to web directory.

Taskfile.yml
fiftin: Fixed

Taskfile.yml
fiftin: Fixed

api/tasks/runner_test.go
twhiston: well it tests if the checkTmpDir will return an error or not if the folder you try to create has a read only parent, as this needs to bubble up to us if we do it by accident. Maybe there is a better way to test it, but we still need to test it, as part of this functions job is to let us know if making the dir was a success or not

Taskfile.yml
twhiston: do they have to be globally installed though? is there a good reason for this? as really they are project level development dependencies

api/tasks/runner_test.go
fiftin: I've replaced OS checking to checking ```os.Chmod``` support. But I still think that this test should be removed because it is not portable.

twhiston: I'll take that for now but in the end it should be fixed so it is portable, because we need to check that our function returns false if we cannot create the directory. 

If you can think of a better and more portable way to check for this then i will happily accept it :smile: basically the condition should be that the function is not allowed to write into the directory that we attempt to create a folder in and this should return an error.

Taskfile.yml
fiftin: https://github.com/pugjs/pug-cli - I'm not found examples of using it locally. 
I tested locally, it works fine, but not documented and looks like hack for me:
```node node_modules/pug-cli resources/pug --out public/html```

**Comment by fiftin** [comment]
```
Found fs.readFileSync with non literal argument at index 0 (security/detect-non-literal-fs-filename)
var contents = fs.readFileSync(file);
```

Too strong check :)

File: Taskfile_windows.yml [hunk 6]
```diff
@@ -0,0 +1,10 @@
+tasks:
```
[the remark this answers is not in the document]
fiftin: > to squash your commits
What is mean? My English is not so good :-))

> ```go get github.com/cespare/reflex || true```
Yes, I fixed it

fiftin:
Looks like Codacy broken after I squashed commits

twhiston:
it just needed to analyze it again, we had some issues yesterday where it got confused between the team and @matejkramny's personal account, he was kind enough to sort it out this morning though so hopefully it is smoother sailing from here on out with codacy reviews!

twhiston:
@strangeman could you give this a quick test as well, it all looks good to me, but as it is a big change to how the web side of the project works it would be cool to have another set of eyes on it (and compiling/running it as a test) before we merge it. Thank you!

fiftin:
I see 1 issue in Codacy for this PR:
```
Found fs.readFileSync with non literal argument at index 0 (security/detect-non-literal-fs-filename)
var contents = fs.readFileSync(file);
```

And severel incorrect issues for FE like this:
```
Found fs.open with non literal argument at index 0 (security/detect-non-literal-fs-filename)
var upgradeModal = $modal.open({
```

Maybe we can disable ```security/detect-non-literal-fs-filename``` check?

twhiston:
> Maybe we can disable security/detect-non-literal-fs-filename check?

Done. Please do keep letting me know of any false positives you see so that we can get codacy configured in the best way possible for how we work. Thanks for this :smile:

strangeman:
@twhiston sure! Will do tomorrow morning.

fiftin:
Thank you, fixed

Issue #140, opened 2016-06-24
Contrib : sample systemd semaphore.service


philippelecoq:
Create file /etc/systemd/system/semaphore.service

---

[Unit]
Description=ansible-semaphore
After=syslog.target
Requires=network.target

[Service]
Type=forking
User=ansible
ExecStart=/bin/sh -c '/usr/bin/semaphore -config  /var/lib/semaphore/semaphore_config.json >> /var/log/semaphore.log 2>&1 &'
Restart=always
RestartSec=8s

[Install]
WantedBy=multi-user.target

---

systemctl enable semaphore
systemctl start semaphore
systemctl status semaphore

matejkramny:
Thanks for this.

This could be added to the setup process. User could pick between `launchctl`, `upstart`, so on..

nwlucas:
I modified the above to be a bit more configurable. This is what I have in use on a staging server:

 /etc/systemd/system/semaphore.service
```shell
[Unit]
Description=Ansible Semaphore
After=syslog.service
Before=nginx.service
Requires=network.target


[Service]
Type=forking
EnvironmentFile=-/etc/default/ansible-semaphore
ExecStart=/bin/sh -c "/usr/bin/semaphore -config ${SEMAPHORE_CONFIG} >> ${SEMAPHORE_LOGS} 2>&1 &"
Restart=always
RestartSec=10s

[Install]
WantedBy=multi-user.target
```


And the accompanying /etc/default/ansible-semaphore
```shell
#Ansible Semaphore Defaults

SEMAPHORE_CONFIG=/data/ansible-semaphore/semaphore_config.json
SEMAPHORE_LOGS=/var/log/semaphore.log

```

philippelecoq:
User=ansible  (or other user name) is very important.
Running ansible with root is not recommended.


Philippe


-------- Message d'origine --------
De : Nigel Williams-Lucas <notifications@github.com> 
Date :27/06/2016  17:27  (GMT+01:00) 
A : ansible-semaphore/semaphore <semaphore@noreply.github.com> 
Cc : "www.MyPiDrone.com" <philippe.le-coq@laposte.net>, Author <author@noreply.github.com> 
Objet : Re: [ansible-semaphore/semaphore] Contrib : sample systemd
  semaphore.service (#140) 

I modified the above to be a bit more configurable. This is what I have in use on a staging server:

/etc/systemd/system/semaphore.service

[Unit]
Description=Ansible Semaphore
[...]

PR #492, opened 2018-03-13
add .goreleaser and associated ci changes

twhiston:
This changes the build and release system to goreleaser.
This causes a number of changes:

- removes build process from taskfile and instead declares it in .goreleaser.yml file
- adds a deploy phase to the ci that will prepare a draft release on tagged commits
- builds .deb and .rpm releases as default 
- removes gox as a dependency, but adds goreleaser
- adds a release:test command which you can use to build all the output files but not create a real release
- build task is now an alias of release:test
- minor fix to circleci v1 file to hardcode the project path instead of taking a dynamic variable so you can build the package properly on circleci when using a fork

@strangeman - as an ops guy it would be really cool if you could take a look at this, install the packages and see that they do what you expect etc.... so we can get this really polished. I'd like to add the signing section (currently commented out) but i'll chat to @Echobob about that as we need to make a decision about what gpg key to use. Additionally in the end i'd also like to include a systemd file with the packaged versions (#140). but we can add that closer to the release of the next version.

I anticipate that there will be some fine tuning to do with the actual release creation when it comes to it, but you need tags and things to really test that so I think we can safely defer it to the actual release time.


.goreleaser.yml
```diff
@@ -0,0 +1,68 @@
...
+      - goos: darwin
+        goarch: 386
+    hooks:
+      pre: task compile
+
+archive:
+  format_overrides:
+    - goos: windows
+      format: zip
+  files:
```
strangeman: Should we put some kind of README (and config.yaml sample) here?

.goreleaser.yml
```diff
@@ -0,0 +1,68 @@
...
+  description: Open Source alternative to Ansible Tower
+  license: MIT
+
+  formats:
+    - deb
+    - rpm
+
+  # Packages your package depends on.
+  dependencies:
+    - ansible
```
strangeman: Better make `ansible` as an optional dependency. Many people use `ansible` from pip.

Taskfile.yml
```diff
@@ -111,17 +112,26 @@ tasks:
...
 
+  release:test:
+    desc: creates a release without performing validations or publishing artifacts
+    cmds:
+      - goreleaser --snapshot --rm-dist
+
+  release:
+    desc: Build a full set of release binaries
+    cmds:
+      - goreleaser
```
strangeman: `goreleaser --rm-dist` is better in case if we make the release on the developer's machine.

```
$ task release
goreleaser

   • releasing using goreleaser dev...
   • loading config file       file=.goreleaser.yml
   • SETTING DEFAULTS FOR:
      • loading environment variables
      • snapshoting              
      • releasing to GitHub      
      • project name             
      • creating archives        
      • building binaries        
      • creating Linux packages with fpm
      • creating Linux packages with nfpm
      • creating Linux packages with snapcraft
      • calculating checksums    
      • signing artifacts        
      • creating Docker images   
      • releasing to Artifactory 
      • creating homebrew formula
      • creating Scoop Manifest  
   • CHECKING ./DIST  
   ⨯ release failed after 0.00s error=bin is not empty, remove it before running goreleaser or use the --rm-dist flag
task: Failed to run task "release": exit status 1
```

strangeman:
I request mainly minor changes. Generally, `gorelease` works great!
In future, we can add features to the packaging (systemd unit, manpage etc), but for now, this is a huge step.

Taskfile.yml
```diff
@@ -42,8 +42,9 @@ tasks:
       - go get -u github.com/golang/dep/cmd/dep
       - go get github.com/cespare/reflex
       - go get -u github.com/gobuffalo/packr/...
-      - go get github.com/mitchellh/gox
       - go get -u github.com/haya14busa/goverage
+      # needs go 1.10+
+      - go get github.com/goreleaser/goreleaser
```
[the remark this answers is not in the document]
twhiston: do you have go 1.10? This works fine for me if i have that, also their install instructions dont give the /...  form. Also try a dep ensure (which seems not needed in the ci but maybe locally where you have downloaded some deps already)

.goreleaser.yml
twhiston: well the readme that is currently in the root kind of says nothing, which is why i didnt include it, its not really very helpful if you want to learn about semaphore. Maybe that is a more general improvement that we could make though.

You can get a config sample by running `semaphore -config` right? The validity of that approach is another question ;) Also you should run setup so you get the database user created, and I feel like adding a config file suggests that if you use it everything will 'just work' which is not the case, and as a new user finding that the missing piece is that you didnt do setup and the user was not created is not actually that easy to find out

Taskfile.yml
twhiston: a real release, with the changelogs etc.... should never be run on a developers machine, that should be only ever done in the ci (imo) as it potentially pushes changes, and it requires a totally clean git tree. In reality it means that only us guys (or the ci) can do it anyway, as no one else is allowed to create releases, but i'm happy to add that flag if you want

Taskfile.yml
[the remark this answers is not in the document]
twhiston: (happy to adjust to your suggestion btw, just that it is not the way in their docs, which is why i didnt do it)

twhiston:
Thanks for the review, i'll change the necessary bits. Definitely want to add more features to the packaging in future so it's as easy to distribute, install and use semaphore as possible, and i'm definitely really open to any suggestions you (or anyone) has about making that happen.

Taskfile.yml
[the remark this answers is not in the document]
strangeman: I tried with Go 1.9 and got this error. Then, I updated to Go 1.10... and still got this error. :)

Taskfile.yml
strangeman: Maybe should add some warning then? Like "NEVER RUN THIS TASK ON DEVELOPER MACHINE" :)

Taskfile.yml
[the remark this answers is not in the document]
twhiston: hahaha, but it was ok with the /... form then? If so i will for sure change it!

Taskfile.yml
twhiston: well it should just fail on a developer machine as they cant push it and the copy will most probably be dirty, so nothing can get totally fucked up from running it, unless one of us forgets about it and runs it. 
However I might remove the description as currently all other commands that people are not meant to call dont have them and as such are hidden from the task list `task -l` or `task` output.
Any better solution you can think of?

Taskfile.yml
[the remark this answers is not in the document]
strangeman: Yup, with /... all was good.

Taskfile.yml
strangeman: >unless one of us forgets about it and runs it
If shit can happen - it will happen. Murphy's law.

Removing description is good for me. Another thought - rename `release:test` to `release` and `release` to `release:prod` (or so). Frequently used commands should be shorter than rarely used (and potentially dangerous) commands.

twhiston: yeah i like that approach. I'll remove description and change the task names to make `release` the safe one

Taskfile.yml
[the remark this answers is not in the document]
twhiston: ok, then it is getting changed. weird that it worked for me and not you, out of interest what OS are you using?

Taskfile.yml
```diff
@@ -42,8 +42,9 @@ tasks:
       - go get -u github.com/golang/dep/cmd/dep
       - go get github.com/cespare/reflex
       - go get -u github.com/gobuffalo/packr/...
-      - go get github.com/mitchellh/gox
       - go get -u github.com/haya14busa/goverage
+      # needs go 1.10+
```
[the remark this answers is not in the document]
twhiston: added

.goreleaser.yml
twhiston: This is the only one I havn't made a change for yet. What do you think we should do?

strangeman: We may put it to separate task, focused on providing of documentation.

Taskfile.yml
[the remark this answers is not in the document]
strangeman: Debian Unstable.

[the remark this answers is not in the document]
twhiston: interesting, im using fedora atomic workstation and building everything inside fedora oci containers

Issue #472, opened 2018-02-28
Use Dredd to test the API


twhiston:
It might we worth looking at using https://github.com/yazgazan/bacom to test the semaphore api and ensure breaking changes are caught in ci

Perhaps could be done at the same time as #464

twhiston:
been looking in to this more and dredd is a better choice as it can validate against the swagger schema. I'll implement this.

twhiston:
done in #524

PR #514, opened 2018-03-27
gometalinter

twhiston:
As making gometalinter happy is going to involve a LOT of small changes to the codebase I want to open this as a w.i.p so that people have a chance to review it in parts. I'll try to make the commits match up with some logic units in the codebase to also make it easier to review.

As this also introduces gometalinter in to the test pipeline the PR is going to have failing tests until every error is either fixed or ignored in the codebase.


twhiston:
first commit fixes (or ignores where appropriate) all issues reported when running `gometalinter --vendor --deadline=120s ./... | grep util/`

twhiston:
Added linting for `cli/main.go` and `db/*` folder.
Updated the linter command to ignore packr files.

twhiston:
Next commit starts to deal with api folder, worked on the sockets and tasks folders. 
This also starts to look at abstracting the error logging out of the api functions so in future we can better use logrus. 
Will squash this commit once I finish the api folder work

twhiston:
So this PR is now done

Summary:
* Solves 262 linter issues
* brings all go files to an A grade in codacy
* checks all errors. This one matters a bit as it will make the log output potentially a bit more noisy, but it's worth it as we should be seeing issues if they happen (and later add the option to report them to external systems via logrus hooks)
* abstracts out some logging. I started to pull the logging of these errors out of the functions via some wrappers, this comes with some heavy caveats at the moment as I only added it in places where we did no checking, as reported by the linter, and its certainly not a fully featured idea yet. I think for now it is ok though, and it saves a lot of code duplication. I'll follow this up with a ticket explaining what i would like to do with the error handling and logging in the long term so we can discuss it and plan it for 2.6.0.

As this makes significant small changes please check that you can still perform all the actions you would expect (we really need a test suite for this stuff to save time and at least validate the api, see #472) and that everything runs properly.

twhiston:
@strangeman on your return please give this a review and let me know if you have any thoughts about the logging

PR #518, opened 2018-04-06
systemd example

twhiston:
example systemd deployment

### DIFF
--- a/deployment/systemd/semaphore.service
+++ b/deployment/systemd/semaphore.service
@@ -1,0 +1,14 @@
+[Unit]
+Description=Ansible Semaphore
+Requires=network.target
+
+[Service]
+EnvironmentFile=/etc/semaphore/env
+ExecStart=/usr/bin/semaphore -config ${SEMAPHORE_CONFIG}
+User=semaphore
+Group=semaphore
+Restart=always
+RestartSec=3s
+
+[Install]
+WantedBy=multi-user.target
--- a/deployment/systemd/util/install.sh
+++ b/deployment/systemd/util/install.sh
@@ -1,0 +1,9 @@
+#!/usr/bin/env bash
+set -e
+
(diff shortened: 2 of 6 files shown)


deployment/systemd/util/install.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+set -e
+
+HERE="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
+mkdir -p /etc/semaphore
+cp ${HERE}/../semaphore.service /etc/systemd/system
```
strangeman: need to call `systemctl daemon-reload` after adding/modifying the systemd unit file.

deployment/systemd/semaphore.service
```diff
@@ -0,0 +1,14 @@
...
+Description=Ansible Semaphore
+Requires=network.target
+
+[Service]
+EnvironmentFile=/etc/semaphore/env
+ExecStart=/usr/bin/semaphore -config ${SEMAPHORE_CONFIG}
+User=semaphore
+Group=semaphore
+Restart=always
+RestartSec=3s
```
strangeman: I think need to set `StartLimitBurst` option, to prevent infinite restarts.
Also, will Semaphore work properly without `WorkingDirectory`?

twhiston: `StartLimitBurst`  will be set to  `DefaultStartLimitBurst` if you dont add anything to the file here (https://www.freedesktop.org/software/systemd/man/systemd.unit.html#), which I thought would be a better default as it is really up to the user how many retries they want. If you would prefer a value let me know what seems sensible to you (we don't set this value in the instances we are running so you probably have a better insight into what is needed)

twhiston: You dont need a working directory, since the place to store ephemeral files (effectively the ansible working directory root) is set via the config

PR #524, opened 2018-04-16
use dredd for api testing

twhiston:
Sorry to drop another fairly large PR into the mix, but I hope you will agree this is worth it.
I've added [dredd](http://dredd.readthedocs.io/en/latest/index.html) to Semaphore as a way to test the api documentation for correctness, and changes leave the api in a working state.
Basically it takes the swagger spec, and tries all the calls against the API, validating that what it gets back matches. At the moment this only tests the 'happy path', but i can imagine adding a few more hooks in future that set up for tests around user access (logged out, admin etc...), or no results responses (some no results responses cause panics so it would also help us to introduce this test when dealing with #462)

I've had to write some hooks for the tests, to make sure that the correct things exist in the database. You should not run these against your production database as the tests produce a tonne of users and projects and doesnt delete them. You can run it against a bare db though, so you can start semaphore, let it run the migrations, you dont even need to create a user, and then you can run dredd, alternatively see the contribution doc for how to run it all in docker. 
Theres an annoying amount of database related code in those hooks, and I realized while doing this that most of the current db related calls are in the endpoints, it would make sense to move that to it's own layer so we could re-use it in the tests etc... I'll try to get to that in 2.6.0

In building out this system I of course found a few things in the API docs that needed changing, so they are also included in this PR. It is quite a bit change but it was mostly stuff like adding explicit request objects (normally with at least the id field missing), moving format strings which showed options to enums, and adding things like min/max values, validation patterns or parameter examples. The parameter examples are important as the values set for those are used later when substituting runtime created objects to manipulate (see  .dredd/hooks/capabilities.go).

On the code side I added 2 small pieces of middleware, 1 that is attached to every route that sets the  content-type header to be json, the other for plain text. This allowed me to normalize all the routes to return json responses as standard, and then override it in routes where necessary (ping).
Additionally to this I caught a small bug whereby return values from new object creation had a slightly different timestamp format than values retried from the database, so I made a normalizer function for that. I additionally added a unit test that the go-openapi library can validate the api schema.

The downside of this is that the tests are a bit slower, i see a pretty easy solution to this though, which is that I will make some base images that we can pull so we don't need the cost of installing all the tools etc.... that should make it significantly faster, especially around dredd, who's compilation time results in  the duration being as long as it is. I'll get that done soon and make a separate issue for it.
I'm also not sure about having the ci stuff in the distributions folder, its not really a distribution, so perhaps it should be moved to be a subfolder of the .circle folder instead?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -7,7 +7,7 @@ aliases:
     image: circleci/golang:1.10
 
   - &working-dir
-      /go/src/github.com/ansible-semaphore/semaphore
+    /go/src/github.com/ansible-semaphore/semaphore
 
   - &store-bin-artifacts
     store_artifacts:
@@ -21,7 +21,7 @@ aliases:
         curl -L https://github.com/go-task/task/releases/download/v2.0.1/task_linux_amd64.tar.gz | tar xvz
         cd -
 
-  - &persist-bin
+  - &persist-from-build
     persist_to_workspace:
       root: .
       paths:
@@ -79,12 +79,12 @@ aliases:
         - v1-go-deps-
 
 jobs:
+
   build:local:
     docker:
       - *golang-image
     working_directory: *working-dir
     steps:
-      - run: export
       - *install-node
       - *install-task-binary
       - checkout
@@ -97,7 +97,7 @@ jobs:
       - *test-compile-changes
       - run: task build:local
       - *store-bin-artifacts
-      - *persist-bin
+      - *persist-from-build
 
   build:
     docker:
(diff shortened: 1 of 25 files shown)


twhiston:
@strangeman Due to the other PR merges I needed to rebase this and make a few small changes. Please give it another quick look before I merge it.

strangeman:
>"technically" its a deployment, but its more like a ci specific deployment that only tests the api, so maybe it should not be there

Hmm. I'm ok with the current location. I think this is not very significant micromanagement :)

PR #531, opened 2018-04-26
Assign users response correctly

### LINKED ISSUE #529: Users page does not work after 2.5.0 upgrade
We replaced our 2.4.1 binary with 2.5.0 and now users no longer populate on the Users page. Have tried this with Admin user with the same result.

Requests to /api/user are only returning the logged in user. No errors in semaphore log and no console errors in chrome tools when browsing.

I looked at /db/migrations/v2.5.0.sql just to make sure we are not requiring a DB change but that does not seem to help.

Any ideas??
users

slickvoodoo:
Fix #529

### DIFF
--- a/web/public/js/controllers/users.js
+++ b/web/public/js/controllers/users.js
@@ -1,7 +1,7 @@
 define(function () {
 	app.registerController('UsersCtrl', ['$scope', '$http', '$uibModal', '$rootScope', function ($scope, $http, $modal, $rootScope) {
 		$http.get('/users').then(function (response) {
-			$scope.users = response.users;
+			$scope.users = response.data;
 		});
 
 		$scope.addUser = function () {


twhiston:
looks like circleci updated their mysql to v8 which has a load of breaking changes, which is causing those tests to fail. You will need most likely need to edit `.circleci/config.yml` to pin to a pre v8 mysql
```
  test:db:migration:
    docker:
      - *golang-image
      - image: circleci/mysql:5.6
```

and also alter `deployment/docker/ci/docker-compose.yml` to pin to that version as well. If you have issues drop me a message on here and i'll pick it up in another pr that you can merge back in to this.

twhiston:
oh and thanks for the fix!

slickvoodoo:
@twhiston Honestly I am not sure how to make the mysql changes. I am not very smart with CI tools or Docker yet :)

twhiston:
Ok no problem, as it doesn't really affect this PR i'm just going to merge it and i'll deal with the tests in a new PR. Thanks for your contribution.

PR #534, opened 2018-05-07
add maintainer information

twhiston:
adds a maintainer to the package to stop warnings on install


twhiston:
tests will fail due to the unpinned mysql versions in the test file. See #536 for the fix to that issues

PR #537, opened 2018-05-08
Fix promise execution, replace then(...).error(...) to then(...).catc…

### LINKED ISSUE #535: New user feature does not work after 2.5.0 upgrade
Not able to create a new user. Below is the error screenshot.

user-creation-error

fiftin:
Fix #535


twhiston:
lgtm

Issue #538, opened 2018-05-08
Bind semaphore on local interface in v2.5.0 does not work


Twim4:
Hi

I am trying to bind semaphore to loopback interface like it was possible in previous v2.4.1 with config entry like:

<pre>
{
        "mysql": {
                "host": "127.0.0.1:3306",
                "user": "xxxxx",
                "pass": "xxxxx",
                "name": "xxxxx"
        },
        "port": "127.0.0.1:3000",
...
</pre>

After semaphore upgrade from 2.4.1 to 2.5.0 the local interface binding is not working anymore:

<pre>
# semaphore -version
v2.5.0
</pre>
after service restart i'm  receiving following error message:

<pre>
Semaphore v2.5.0
Port :127.0.0.1:3000
MySQL semaphore@127.0.0.1:3306 semaphore
Tmp Path (projects home) /opt/semaphore
Checking DB migrations
time="2018-05-08T15:52:22+02:00" level=panic msg="listen tcp: address :127.0.0.1:3000: too many colons in address"
panic: (*logrus.Entry) (0x9eefc0,0xc42027e5a0)
</pre>

Maybe it would be a good idea to implement such parameter like "(bind_)ip": "0.0.0.0" or something similar.

Does somebody know some workaround for this issue to be able to bind it to local interface?

Thanks!

twhiston:
Technically the fact you used to be able to put a full IP in there is an abuse of the field and we added validation for the port (because it should only be a port) to ensure it starts with `:` due to error reports where people expected that you could just put the number in the field.

So in your example it is prepending `:` to your string see `util/config.go` L208.
If we want to allow a full ip here we should introduce another config field for it. @strangeman thoughts?

Twim4:
totally agree with explanation about the port config line.

But the thing is here, that i want to hide semaphore behind nginx reverse proxy, to not be able to connect to it from elsewhere except proxy. And with this config it is not possible.

twhiston:
Your use case is totally valid and i'm happy to support it, but we will need to add a new field rather than revert the port field fix. 
As a temp workaround you can always compile it without that check as we only do it in one place (config.go L197), but in the long term we need a proper solution. It would actually be really simple as we only use the port in a few places. I would of course be very happy to accept a PR if you want to make one.

PR #540, opened 2018-05-14
add interface variable to settings

twhiston:
fix for #538.
I guess this should be in 2.6.0 since adding a variable is not a patch action

### DIFF
--- a/cli/main.go
+++ b/cli/main.go
@@ -36,6 +36,7 @@ func main() {
 	}
 
 	fmt.Printf("Semaphore %v\n", util.Version)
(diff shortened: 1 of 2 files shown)


markfaine:
Has this been merged into any release that is currently available for download?  I'm having the same issue.  I'm also behind an nginx proxy in a docker container and I can't force it to bind to my overlay network.  This means it's unavailable outside of the container because it listens on 0.0.0.0 and in this case that is 10.0.0.2

fiftin:
@markfaine Done: https://github.com/ansible-semaphore/semaphore/releases/tag/v2.5.2

PR #584, opened 2019-07-09
Set version on windows taskfile

megakoresh:
It doesn't otherwise build on windows since by default it's `1`: 
```powershell
Taskfiles versions should match. First is "2" but second is "1"
```


megakoresh:
Superseded by https://github.com/ansible-semaphore/semaphore/pull/585

PR #585, opened 2019-07-09
Fix mulekick dependency removal

megakoresh:
Removal of mulekick dependency has broken the whole project. (that time I said it "builds" fine for me? Yeah, it didn't, I confused "compile" with "build"). This should fix it by changing the code to use gorilla/mux router directly. Mulekick has been effectively removed. 

Two functions that remained have been moved to util package. `router.go` contains most of the changes. Looks a bit uglier in places, but hopefully more robust and transparent now.

With my node.js background I tried to make mux behave like Koa first, which is not a bright idea of course. Still - could be useful, so I kept it in history.


megakoresh:
There is some issue with CI setup, specifically with go-task, but I don't know how to resolve that. Anyway I have built and tested it also manually by just clicking around in the UI, nothing appears to be broken, at least based on what I have seen. And at any rate it builds fine, which is better than current state of upstream. Maybe @matejkramny can take a look and merge this?

matejkramny:
@strangeman any comments? Looks good to me.

One thing i'm noticing (unrelated to the PR), the `gorilla/context` package is deprecated in favour of `r.Context`. In other projects it was causing me severe memory leaks.

megakoresh:
> @strangeman any comments? Looks good to me.

I guess the guy has left somewhere. Can you merge this? As much potential as I see with this project, it doesn't give a strong impression with stagnation like this. Besides - he merged the PR that removed mulekick initially, which broke the upstream completely, it can't get worse if you merge a fix to that.

Frankly to get more activity on the project, requirements to merge stuff need to be loosen and speed of merging and closing issues increased. Not enough people are active on, or using this software for that to be a risk at this stage anyway. A bigger risk is the seeming stagnation of the community, since despite all the potential this project has as an alternative to the (very faulty) AWX/Tower, it won't be a serious contender to it, if it doesn't get more popular and gain a more active community.

matejkramny:
fair point @megakoresh. I'll take that into consideration.

PR #608, opened 2020-10-01
Support SSL/TLS connection to MySQL

fiftin:
https://github.com/ansible-semaphore/semaphore/issues/559

Docs: https://github.com/go-sql-driver/mysql#tls


fiftin:
Not working

PR #634, opened 2020-12-01
Get rid of MySQL dependency

fiftin:


fiftin:
Issue: https://github.com/ansible-semaphore/semaphore/issues/238

PR #653, opened 2021-03-25
fix: The drop-down menu is too long to scroll

bafuka:


bafuka:
image

The drop-down menu is too long, but it cannot be scrolled causing the following items that cannot be seen.

bafuka:
Issues #654

Issue #699, opened 2021-08-27
Add possibility to cancel popups via ESC button


POPSuL:
All of the popup dialogs have persistent vue prop, which prevent close popup by click in outside of dialog and prevent dialog canceling via ESC button.
That is very inconvenient in terms of UX.

fiftin:
@POPSuL Merged, thank you

PR #700, opened 2021-08-27
Allow to close popups via ESC button

POPSuL:
For #699


fiftin:
ESC works only after popup focused. Do you know how to fix it?

POPSuL:
@fiftin hmmm... It can be fixed by adding something like that:
```js
beforeCreate() {
  window.addEventListener('keydown', (ev) => {
    if (ev.key === 'Escape' && this.dialog !== false) {
      this.close();
    }
  });
},
```
into `EditDialog` instead of `@keydown.esc` handler.
Make a pull request?

PR #707, opened 2021-08-30
Added global ESC handler to close popups

POPSuL:
Related to https://github.com/ansible-semaphore/semaphore/pull/700#issuecomment-908095390

### DIFF
--- a/web2/src/components/EditDialog.vue
+++ b/web2/src/components/EditDialog.vue
@@ -10,7 +10,6 @@ Can use used in tandem with ItemFormBase.js. See KeyForm.vue for example.
     persistent
     :transition="false"
     :content-class="'item-dialog item-dialog--' + position"
-    @keydown.esc="close()"
   >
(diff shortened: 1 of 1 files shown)


web2/src/components/EditDialog.vue
```diff
@@ -81,6 +80,14 @@ export default {
...
   },
 
+  beforeCreate() {
+    window.addEventListener('keydown', (ev) => {
+      if (ev.key === 'Escape' && this.dialog !== false) {
+        this.close();
+      }
+    });
+  },
+
```
fiftin: I think it is bad idea. This listener must be removed manually when form closed.

POPSuL: Ok, I'll try rewrite to mounted/unmounted

POPSuL: Rewritten to watcher

PR #743, opened 2021-09-22
Secure email

wowi42:
As part of one deployment, we need to use Mailgun to send Emails.

We needed:

- username
- password
- starttls

Everything is in the PR, Let me know if you need more details!


fiftin:
Thank you!
Changes will be included in the next release.

PR #753, opened 2021-10-03
Audit Logs (HTTP Webhooks)

wowi42:
The idea behind this feature is to have more notifications to 3rd party services!

Instead of writing one connector to each service, we can easily send an HTTP POST with the following body:

```
{
                "start":      t.task.Start,
                "end":        t.task.End,
                "status":     t.task.Status,
                "task_id":    t.task.ID,
                "project_id": t.projectID,
                "task_url":   util.Config.WebHost + "/project/" + strconv.Itoa(t.template.ProjectID)
}
```

Based on this post, you could easily parse it and plug it inside IFTTT, n8n, or Zapier.

It's directly integrated inside the ```updateStatus()``` function. so we should not miss any events

Feedbacks most welcome!.


wowi42:
More fields have been added, and I have  build the complete task execution URL inside the answer:

```
{
    "end": null,
    "environment": "",
    "playbook": "",
    "project_id": 1,
    "start": "2021-10-05T00:19:49.557209331+04:00",
    "status": "running",
    "task_id": 2147483638,
    "task_url": "/project/1/history/?t=2147483638",
    "template_id": 1
}
```

wowi42:
@fiftin Any feedback?

PR #760, opened 2021-10-13
forcing local repo checks for 'file://' prefixed repositories - issues/715

jbkc85:


fiftin:
Thank you @jbkc85 !

jbkc85:
> Thank you @jbkc85 !

not a problem @fiftin - I hope there was a way to test this out...I still have yet to be able to get all the javascript stuff up and running on my local machine :-(

api/tasks/runner.go
```diff
@@ -173,10 +175,17 @@ func (t *task) prepareRun() {
 		return
 	}
 
-	if err := t.updateRepository(); err != nil {
-		t.log("Failed updating repository: " + err.Error())
-		t.fail()
-		return
+	if strings.HasPrefix(t.repository.GitURL, gitURLFilePrefix) {
+		repositoryPath := strings.TrimPrefix(gitURLFilePrefix, t.repository.GitURL)
```
cameronpm, line 203: Wrong way round, the prefix is the 2nd argument

https://pkg.go.dev/strings#TrimPrefix

api/tasks/runner.go
```diff
@@ -173,10 +175,17 @@ func (t *task) prepareRun() {
...
+		if _, err := os.Stat(repositoryPath); err != nil {
+			t.log("Failed in finding static repository at " + repositoryPath + ": " + err.Error())
```
cameronpm, line 205: Should fail() and return here ?

api/tasks/runner.go
jbkc85: ugh, seriously?  I thought I had that appropriately set.

api/tasks/runner.go
jbkc85: Yes.  It should.

PR #811, opened 2022-01-11
Added ansible limit to API

jdhayes:


jdhayes:
Added function "--limit" to API, in order to specify the calling host.

fiftin:
Hi @jdhayes 

What reason to add `deployment/docker/ci/Singularity` and `deployment/docker/ci/bin/build.sh` ?

jdhayes:
Both files are used to install semaphore within a singularity container rather than a docker one.
Most users would not fond these useful, however within HPC environments running Docker is prohibited, but Singularity is possible.

Issue #756, opened 2021-10-09
So does this support LDAPS or not?


KenK73:
I see there is an "ldap_needtls" boolean in the config, but I cant find any mention of how to make it work.  My Semaphore works fine under normal LDAP, but where I want to use it, straight LDAP is not permitted and LDAPS must be used.

If I add the "needtls: true" and change the port to 636, I get this error in the log when attempting to log in:

`Checking DB migrations`
`MySQL root@172.17.0.2:3306 semaphore`
`Tmp Path (projects home) /tmp/semaphore`
`Semaphore v2.7.22`
`Interface`
`Port :3000`
`Server is running`
`INFO[0013] unable to read LDAP response packet: read tcp 172.17.0.11:34666->10.0.30.135:636: read: connection reset by peer`

This is a docker container, so I copied my root/intermediate CA certs into /etc/ssl/certs/ca-certificates.crt but that had no impact.

fiftin:
I adding support LDAPS but I need help with testing.
I will create pre-release soon.

KenK73:
Thanks!  I will assist in the testing as best I can.

Bkbest:
Hi @fiftin, I will also test this.

fiftin:
Hi @KenK73, @Bkbest
Please try v2.8.0: https://github.com/ansible-semaphore/semaphore/releases/tag/v2.8.0

You need to set flag `ldap_needtls: true` in config file before.

KenK73:
I am getting this error now:

Semaphore v2.8.0
Interface
Port :3000
Server is running
INFO[0023] LDAP Result Code 200 "Network Error": ldap: already encrypted
INFO[0973] LDAP Result Code 200 "Network Error": ldap: already encrypted

Here are my ldaps settings:

        "ldap_needtls": true,
        "ldap_binddn": "cn=svc.bind.semaphore,ou=service,ou=users,ou=test,dc=test,dc=lab",
        "ldap_bindpassword": "password",
        "ldap_server": "mad01.test.lab:636",
        "ldap_searchdn": "ou=users,ou=test,dc=test,dc=lab",
        "ldap_searchfilter": "(&(objectCategory=Person)(sAMAccountName=%s))",
        "ldap_mappings": {
                "dn": "dn",
                "mail": "mail",
                "uid": "sAMAccountName",
                "cn": "cn"
        },
        "ldap_enable": true,

[...]

PR #854, opened 2022-02-04
Ldap fixes related to #756

srinathman:
Please consider these changes/fixes to LDAP authentication:

- adding ldap_skip_verify_certs config parameter
- Using new DialURL function for ldap connections
- removed ldap_needtls config in favor of ldap_starttls. This tells the app to use StartTLS upgrade (TLS over regular 389 port)
- make authenticated binding to ldap server optional

**_ldap_server_** can use URLs like:

- ldap://<server>  (default port: 389)
- ldap://<server>:<port>
- ldaps://<server> (default port: 636)  
- ldaps://<server>:<port>

if **_ldap_starttls_** is set to true, it will try to upgrade plain text connections to TLS using StartTLS

if **_ldap_skip_verify_certs_** is set to true, remote certificates will not be verified


gefeng24:
这是来自QQ邮箱的假期自动回复邮件。
 
您好，我一看到您的邮件就会第一时间给您回复。

Issue #933, opened 2022-05-16
Document customizing the Semaphore container(s)


jcmdln:
For my use of Semaphore, I need netaddr installed for several playbooks but it's missing in the official container images. Looking around I saw some building totally new images, though should Semaphore briefly document how to easily customize the official image? I'd like to discuss the value and framing of providing something like the following example of adding py3-netaddr:

```Dockerfile
FROM ansiblesemaphore/semaphore

# Set user to 'root'
USER root

# Install additional system dependencies
RUN apk --no-cache add py3-netaddr

# Set user back to 'semaphore'
USER semaphore
```

dgibbs64:
I also need netaddr installed in the docker container. I think it would be worth just getting it added in to the official image

Apollo3zehn:
I would like to have rsync installed to be able to use the `ansible.posix.synchronize` task. For now I am installing it manually using `sudo podman exec -it -u root semaphore apk add rsync` which is fine for now.

But my first approach was to create a play which does the installation of that package. However, no `sudo` or `doas` is available in the container so I could not get sudo permissions to install the package.

PR #934, opened 2022-05-18
docker: bump alpine from 3.13 -> 3.15

jcmdln:
How do you feel about bumping the golang and Alpine versions? Having access to a newer Alpine makes it simpler to add Ansible dependencies to Semaphore containers as I brought up in #933 and this would help with being able to concisely extend the Semaphore containers.


jcmdln:
Actually, bumping either version needs far more work than this. I'll submit a new PR with my local changes.

PR #935, opened 2022-05-18
Run gofmt on all '*.go' files

jcmdln:
Is there a reason for not using gofmt? I think it's really valuable for readability across projects.


jcmdln:
Actually, bumping either version needs far more work than this. I'll submit a new PR with my local changes.

fiftin:
Thank you @jcmdln , will do it later.

PR #936, opened 2022-05-18
deployment: rework and update prod image

jcmdln:
I'd like to update the Semaphore container images, starting with `prod` as an example, so that bundling additional dependencies into Semaphore as I described in #933 can be a little easier. 

* Drop Python 2.x requirement from `node-gyp` by bumping various dependencies
* Use ansible-core instead of ansible package, as collections/roles should be in Galaxy requirements files
* Include additional dependencies such as py3-netaddr, which is required by various modules and filters for IP parsing

The primary caveat to these changes is that `npm audit` is extremely unhappy, but I'm not familiar enough with Vue to understand what _should_ occur past bumping only the dependencies that are required in order to build and run Semaphore. I'd like to get feedback on how to proceed with this.


fiftin:
Hi @jcmdln ,

Did you test this change?

jcmdln:
I haven't tested using dredd, that would involve rewiring the CI and Dev containers as well. I plan to do so but first wanted to get feedback on my approach, and get clarification on how the Vue dependencies are managed/maintained.

fiftin:
Hi @jcmdln 

I updated vue deps today.

PR #962, opened 2022-06-14
Add back ENV support to environment

dpskvn:
This is basically a re-implementation of https://github.com/ansible-semaphore/semaphore/pull/458 which was dropped in version 2, I believe.

As discussed here: https://github.com/ansible-semaphore/semaphore/issues/951, this would cover the scenarios where the dynamic inventory requires a token passed as an environment variable and if you have several accounts (which have different tokens), hard-coding it in the unit file is not an option.

I'm open to suggestions/comments. I know the contributing guidelines say there should be an issue opened first, but there is already one which I mentioned in the previous paragraph and there have been no replies to it.


fiftin:
Hi @dpskvn,

I sure we should have another place for environment variables. Perhaps additional textarea for it.

dpskvn:
@fiftin  The environmental variables are a now a separate text area inside the Environment section. The new database model also reflects that.

fiftin:
Thank you @dpskvn , LGFM, but after merge I will make refactoring to support YAML format instead of JSON (I will rename `json` to `extra_vars`, `env` to `env_vars`).

PR #965, opened 2022-06-15
Email: Task success alert

YoungjuWang:


AndrewSav:
You are removing the option of sending a success alert to telegram while not sending it to email. I'd like that option to be retained.

fiftin:
Hi @YoungjuWang ,

I agree with @AndrewSav. Email in my mind used for critical alerts only. I don't want have a spam to email, but I wan't to see all events in telegram.

Perhaps we need appropriate flag for each alert destination.

YoungjuWang:
Thank you for feedback

Issue #998, opened 2022-07-22
Execute task-template within python-virtual environment or container


ccuz:
The task-template are run by https://github.com/ansible-semaphore/semaphore/blob/develop/services/tasks/runner.go that calls https://github.com/ansible-semaphore/semaphore/blob/develop/lib/AnsiblePlaybook.go which starts the ansible-playbook command into a sub-process.

We have some shortcomings with this simple effective execution:
- Using '-vvv' while execution the ansible script displays the environment variable that started semaphore (i.e. db password, token, ... if configured through the semaphore-wrapper using environment variable).
- Per project, we use specific version of ansible and collections (using roles/requirements.yaml and collections/requirements.yaml). That works, though we also patches these collections using git-patches.

One idea, would be always build a python venv (https://docs.python.org/3/library/venv.html) before running the task-template process, thus dependencies doesn't clashes.

Another would be to create a container (i.e. docker image) an run the ansible within the provided container image. This provides a solution for problem 2 and ensure no semaphore environment variable leeks into the ansible-playbook execution. Not sure if possible in go while keeping everything simple, maybe using https://www.redhat.com/sysadmin/podman-inside-container. This would match https://access.redhat.com/documentation/en-us/red_hat_ansible_automation_platform/2.1/html-single/red_hat_ansible_automation_platform_upgrade_and_migration_guide/index#upgrading-to-ees

Maybe, the more flexible would be to allow a '/etc/semaphore/playground-setup.sh' to be run just before the ansible-playbook command, controlled by semaphore admin. Thus one could spin a container and pass the 'ansible-playbook' command and argument to it or setup a python venv before each execution. Or just use git hooks https://schacon.github.io/git/githooks.html#_post_checkout to perform environment setup (i.e. patching collections or setting up python venv).

ccuz:
I ended up creating a `hooks/post-checkout` and `hooks/post-merge` within the ansible git project:
```
#!/bin/bash
echo "Git repo hook: $0"

echo "Create a python venv"
python3 -m .venv myproject
source .venv/bin/activate
pip install -r pip-packages.txt

echo "Installing roles/requirements.yml"
ansible-galaxy role install -r roles/requirements.yml --force

echo "Installing collections/requirements.yml"
ansible-galaxy collection install -r collections/requirements.yml --force

echo "Patching buggy collections"
git apply $(ls patches/*.patch)
``` 

In the dockerfile where semaphore is installed, I added:
[...]

PR #1001, opened 2022-07-26
#998 Execute task-template within python-virtual environment and ensu…

ccuz:
…re sensitive environment variable are not passed to ansible-playbook command

### DIFF
--- a/lib/AnsiblePlaybook.go
+++ b/lib/AnsiblePlaybook.go
@@ -16,10 +16,23 @@ type AnsiblePlaybook struct {
 }
 
 func (p AnsiblePlaybook) makeCmd(command string, args []string, environmentVars *[]string) *exec.Cmd {
-	cmd := exec.Command(command, args...) //nolint: gas
+	commandToExec := command
+	cmdInPythonDefaultVenv := fmt.Sprintf("%s/.venv/bin/%s", p.GetFullPath(), command)
+	if _, err := os.Stat(cmdInPythonDefaultVenv); !os.IsNotExist(err) {
+		// Run .venv/bin/command instead of the one in PATH
+		commandToExec = cmdInPythonDefaultVenv
+	}
+	cmd := exec.Command(commandToExec, args...) //nolint: gas
 	cmd.Dir = p.GetFullPath()
 
 	cmd.Env = os.Environ()
+	pythonDefaultVenv := fmt.Sprintf("%s/.venv", cmd.Dir)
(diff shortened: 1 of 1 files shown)


ccuz:
Based on idea #998 this implementation allow to run the semaphore ansible task within a prepared python virtual environment '.venv' (default naming per convention for python venv)
grafik.

Note: In order to prepare the venv, one still has to use git post-checkout and post-merge hooks https://github.com/ansible-semaphore/semaphore/issues/998#issuecomment-1192731622. This is not covered by any pull-request as it works as is from semaphore git checkout

fiftin:
Hi @ccuz, sorry for delay. I reviewing your PR, thank you!

fiftin:
It looks very interesting.

ff-fgomez:
Is there planned support for multiple venvs? It would be a great feature to add

lib/AnsiblePlaybook.go
```diff
@@ -16,17 +16,36 @@ type AnsiblePlaybook struct {
 }
 
 func (p AnsiblePlaybook) makeCmd(command string, args []string, environmentVars *[]string) *exec.Cmd {
-	cmd := exec.Command(command, args...) //nolint: gas
+	commandToExec := command
+	cmdInPythonDefaultVenv := fmt.Sprintf("%s/.venv/bin/%s", p.GetFullPath(), command)
```
andreas-marschke, line 20: This is unlikely to work nicely with every Ansible setup since playbooks could be nested into subdirectories of the ansible project. 
Consider the following project setup: 

```
.       # Project Root
roles/
roles/roleA
roles/roleB
roles/roleC

playbooks/
playbooks/projectA
playbooks/projectA/playbookA.yml
playbooks/projectB/playbookB.yml
```
If your task template would reference `playbooks/projectA/playbookA.yml` then `p.GetFullPath()` may now resolve to (assuming other things being default) `/tmp/semaphore/project_1/playbooks/projectA/`, which will not be where your .venv would be. 

Instead I'd recommend having the local venv dependencies deployed into the `$PATH` through the either activating it prior to execution or into the env that semaphore is ultimately running in.

My current installation(s) use the core semaphore container but install addtl. dependencies into the container.

Yes, this requires to update the container if you have a new dependency that needs to be installed but it does get around these presumptive and rather brittle constructs.

andreas-marschke (changes requested):
I want to agree with the notion of cleaning up/removing ENV vars that could leak information, but I'd prefer it if that concern got separated from the virtualenv work.

PR #1056, opened 2022-10-14
Using docker secrets to pass sensitive information via "_FILE" variables.

vaerh:
Hi Denis, can you review this patch?
We would like to remove all sensitive information from the environment variable file and use docker secrets to pass it.

### DIFF
--- a/deployment/docker/common/semaphore-wrapper
+++ b/deployment/docker/common/semaphore-wrapper
@@ -4,6 +4,26 @@ set -e
 
 echoerr() { printf "%s\n" "$*" >&2; }
 
+file_env() {
+        local var=""
+        local fileVar=""
+        eval var="\$${1}"
+        eval fileVar="\$${1}_FILE"
+        local def="${2:-}"
+        if [ -n "${var:-}" ] && [ -n "${fileVar:-}" ]; then
+                echo >&2 "error: both ${1} and ${1}_FILE are set (but are exclusive)"
+                exit 1
+        fi
+        local val="$def"
+        if [ -n "${var:-}" ]; then
+                val="${var}"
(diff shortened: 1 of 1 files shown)


fiftin:
Hi @vaerh ,

It is very useful improvement. Сan it break the old behavior?

vaerh:
I hope it doesn't break. This patch is made to look like a working implementation for [mariadb-docker](https://github.com/MariaDB/mariadb-docker/blob/master/docker-entrypoint.sh#L21-L40)
And we have the production stack with these changes.

fiftin:
Do you have any docs for this feature?

vaerh:
I'm sorry to be late with the reply.
Maybe not invent anything and slightly correct the excellent description from the [MySQL Team](https://hub.docker.com/_/mysql/)?

**Docker Secrets**
As an alternative to passing sensitive information via environment variables, _FILE may be appended to the previously listed environment variables, causing the initialization script to load the values for those variables from files present in the container. In particular, this can be used to load passwords from Docker secrets stored in /run/secrets/<secret_name> files. For example  ```docker-compose.yml```:



```yaml
semaphore:
    restart: unless-stopped
    ports:
      - 3000:3000
    image: semaphoreui/semaphore:latest
    environment:
      SEMAPHORE_DB_PASS: /run/secrets/AnsibleSemaphoreMySQLPassword
```
Currently, this is only supported for ```SEMAPHORE_DB_USER```, ```SEMAPHORE_DB_PASS```, ```SEMAPHORE_ADMIN```, ```SEMAPHORE_ADMIN_PASSWORD```, ```SEMAPHORE_LDAP_PASSWORD``` and ```SEMAPHORE_ACCESS_KEY_ENCRYPTION```.

Issue #1068, opened 2022-10-28
No collections/requirements.yml file found.


VladoPortos:
This is quite confusing.

I have the folder and file in the git next to the yaml file that is called.

collections/requirements.yml 

```
---
collections:
  - name: ansible.posix
```

```
/tmp/semaphore/repository_1_1 $ ls -all
total 28
drwxr-xr-x    4 semaphor root          4096 Oct 28 16:39 .
drwxr-xr-x    1 semaphor root          4096 Oct 28 16:34 ..
drwxr-xr-x    8 semaphor root          4096 Oct 28 16:34 .git
-rw-r--r--    1 semaphor root          6240 Oct 28 15:48 README.md
-rw-r--r--    1 semaphor root           563 Oct 28 16:18 add_user.yaml
drwxr-xr-x    2 semaphor root          4096 Oct 28 16:30 collections
/tmp/semaphore/repository_1_1 $ ls collections/
requirements.yml
```

Why is semaphore not seeing it ?  Its the last thing that I need to figure out before I can use semaphore.

VladoPortos:
Well unless I'm doing something wrong this just don't work... I tried in docker version and the latest .deb on dedicated VM... same result... collections are ignored roles are detected

VladoPortos:
Another interesting stuff:
 Just for lolz I moved the collection folder into /playbook directory . This one is defined here: 
```
"tmp_path": "/playbook",
```

And then it finally found the file... it is looking one folder off. Although checking the code it should see it in the git folder...

```
requirementsFilePath := path.Join(t.getPlaybookDir(), "collections", "requirements.yml")
```

Maybe this change is not present in v2.8.73 ?

VladoPortos:
Checked the release code of v2.8.73 and its there... I do not know go well enough to debug this... I tried one last thing and thats to use default path "/tmp/semaphore/" but that does not help either.... this is beyon me :-/

fiftin:
Hi @VladoPortos 
Thank you very much, it is regression! Will be fixed ASAP.

VladoPortos:
Looks to be working in the latest release, thanks ! Enjoy the coffee :)

fiftin:
@VladoPortos Thank you :) Fixed in [v2.8.74](https://github.com/ansible-semaphore/semaphore/releases/tag/v2.8.74).

bverkron:
I seem to be getting the same issue using the latest docker image (2.8.74, I see the new template UI changes)

[...]

PR #1112, opened 2022-12-07
fix(runner): load {roles,collections} requirements files from repo root like AWX #1068

Zaephor:
Got bitten by the differences highlighted https://github.com/ansible-semaphore/semaphore/issues/1068#issuecomment-1303042950.

I know this isn't AWX, but following a similar expected structure may help with adoption by both AWX users and anyone with too many playbooks to keep them in the root of the repository.

This could technically be a breaking change for some users depending on their repository structured. I'm not sure if you would prefer to consider supporting multiple requirements files to support both AWX path logic and the existing logic.

### DIFF
--- a/services/tasks/runner.go
+++ b/services/tasks/runner.go
@@ -502,7 +502,7 @@ func (t *TaskRunner) updateRepository() error {
 }
 
 func (t *TaskRunner) installCollectionsRequirements() error {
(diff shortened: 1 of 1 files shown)


tntclaus:
What prevents this PR from being merged?

piyoki:
This feature is very necessary. I am experiencing exactly the same issue!

piyoki:
@fiftin Could we get some review on this, please? Context linked to #1068

PR #1178, opened 2023-03-07
Added go-git client support

mgukov:
Added go-git client support and switched to use it by default.


fiftin:
Hi @mgukov thank you!)

PR #1209, opened 2023-04-10
add tzdata and environment variable

BadCo-NZ:
added tzdata package and TZ environment variable

untested yet :smile:


tboerger:
This PR can be get closed as this is already integrated.

PR #1231, opened 2023-05-04
Don't `--force` install collections

jcmdln:
If we already have the referenced collections/roles then there is no need to re-download them. If users require specific versions this can be specified in the associated requirements files which would trigger a download as needed.

### DIFF
--- a/services/tasks/runner.go
+++ b/services/tasks/runner.go
@@ -519,7 +519,6 @@ func (t *TaskRunner) installCollectionsRequirements() error {
 			"install",
 			"-r",
 			requirementsFilePath,
-			"--force",
 		}); err != nil {
 			return err
 		}


Thulium-Drake:
I don't agree, that's not how galaxy works, unfortunately... If you want to make _any_ change to installed roles/collections, you must use ```--force```. Even when doing updates, as Galaxy client does not yet provide a facility for this task.

jcmdln:
If you have a `requirements.yml` file with the following format without a version specifier, the collections will be installed if missing:

```
# requirements.yml
---
collections:
  - name: community.crypto
  - name: community.general
```

```
$ ansible-galaxy install -r requirements.yml 
Starting galaxy collection install process
Nothing to do. All requested collections are already installed. If you want to reinstall them, consider using `--force`.
```

If I then specify a version for a single collection, ansible-galaxy triggers the install of the specified version:

```
---
collections:
  - name: community.crypto
  - name: community.general
    version: "6.5.0"
```

```
$ ansible-galaxy install -r requirements.yml 
Starting galaxy collection install process
Process install dependency map
Starting collection install process
Downloading https://galaxy.ansible.com/download/community-general-6.5.0.tar.gz to /home/user/.ansible/tmp/ansible-local-26013207r21uwhr/tmpfs57mfrl/community-general-6.5.0-xzox02d5
'community.crypto:2.13.0' is already installed, skipping.
Installing 'community.general:6.5.0' to '/home/user/.ansible/collections/ansible_collections/community/general'
community.general:6.5.0 was installed successfully
```

grvlbit:
I think this discussion needs to be separated into roles and collections.

For collections there is a `--upgrade` flag for `ansible-galaxy collection install` (see https://docs.ansible.com/ansible/latest/collections_guide/collections_installing.html#installing-collections-with-ansible-galaxy). 
However, as @Thulium-Drake already pointed out this won't work for roles. For roles the upgrade feature has been requested for a long time [ansible/ansible issue 6466](https://github.com/ansible/ansible/issues/6466) but hasn't been implemented so far.

My suggestion is to replace the `--force` flag by the `--upgrade` flag for collections and leave the roles install unchanged until the upgrade feature is implemented upstream at some point.

jcmdln:
> My suggestion is to replace the `--force` flag by the `--upgrade` flag for collections and leave the roles install unchanged until the upgrade feature is implemented upstream at some point.

I agree with this.

services/tasks/runner.go
```diff
@@ -519,6 +519,7 @@ func (t *TaskRunner) installCollectionsRequirements() error {
 			"install",
 			"-r",
 			requirementsFilePath,
+			"--update",
```
NiceRath: It should be '--upgrade' not '--update'!

jcmdln:
@NiceRath Thanks!

After some extended use, personally I don't want the globally installed collections to be upgraded on every run as this causes a lot of churn and API calls to Ansible Galaxy, especially with large and complicated playbooks. Existing Collections should be left alone unless the `requirements.yml` specifies a version that isn't installed.

NiceRath:
Im my opinion it would make sense that users can configure this behaviour themselfs on [repository scope](https://docs.ansible-semaphore.com/user-guide/repositories).

For example:
- Requirements
  - ( ) Ignore (manually managed)
  - ( ) Install
  - ( ) Upgrade (role: force, col: upgrade)
  - ( ) Force

jcmdln:
My goal with this PR was to avoid Ansible Galaxy timeouts with the current behavior. Since I appear to have a unique constraint I'll defer to carrying this as a patch for my own needs as to not interrupt the flow of others.

00willo:
Linking to the ansible role update feature proposal (https://github.com/ansible/proposals/issues/23) for completeness.

PR #1244, opened 2023-05-18
bug fix, issue #1247 

### LINKED ISSUE #1247: [BUG] list of users stops at 20, no button for "next page"
Hi

**semaphore version 2.8.77**

when you have more than 20 users registered, in the Users list at the bottom, you can't see past 20 users and you can't change page, there is not even a button for it.

this is quite annoying and breaking functionality in an enterprise environment.

thanks

battlecook:
add trim for RepositoryForm.vue
bug fix https://github.com/ansible-semaphore/semaphore/issues/1247

### DIFF
--- a/web/src/views/Users.vue
+++ b/web/src/views/Users.vue
@@ -44,9 +44,9 @@
     <v-data-table
       :headers="headers"
       :items="items"
-      hide-default-footer
-      class="mt-4"
       :footer-props="{ itemsPerPageOptions: [20] }"
+      :hide-default-footer="hideFooter"
+      class="mt-4"
     >
       <template v-slot:item.external="{ item }">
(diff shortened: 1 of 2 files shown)


web/src/views/Users.vue
```diff
@@ -44,9 +44,9 @@
     <v-data-table
       :headers="headers"
       :items="items"
-      hide-default-footer
-      class="mt-4"
       :footer-props="{ itemsPerPageOptions: [20] }"
+      :hide-default-footer="hideFooter"
```
fiftin, line 48: No field `hideFooter` in Users component.

fiftin: May be `hide-default-footer` should be romoved?

fiftin:
Fixed in other PR

PR #1250, opened 2023-05-27
Enhancement: Installation of User Python Dependencies at Container Runtime

### LINKED ISSUE #1245: Enhancement: Installation of User Python Dependencies at Container Runtime
Hello! I would like to propose an enhancement to the docker deployment to support installation of additional python packages for use with Ansible. It appears that this can be done when Semaphore is installed without docker (where the host system provides the Ansible installation), but this capability is not available when deploying with docker. I am able to submit a PR for this change, but wanted to get feedback first to make sure it would be accepted.

Proposed Change:
---
I propose to add logic to `deployment/docker/common/semaphore-wrapper` to look for a file called `requirements.txt` within the `$SEMAPHORE_CONFIG_PATH` directory. If present, the script would invoke `python -m pip install -r "${SEMAPHORE_CONFIG_PATH}/requirements.txt"` to install the requirements.

Advantages:
---
- This solution has no breaking changes since the absence of a requirements.txt file would cause the script to behave as it does today. No logic has to be added to the Semaphore app code.
- This allows users to take advantage of a wider variety of Ansible collections that rely on additional python packages.
- We can take advantage of the existing pip requirements file format which is well documented.

Disadvantages:
---
- Additional startup time for the container while packages are installed. For a few packages, this won't be a huge deal, but for larger requirements files this could add significant time.
- Packages would be installed every time the container starts. Again, this wouldn't matter too much for a few packages, but doesn't scale well.

Alternate Solutions:
---
- Could add support for an additional, optional data volume to be mounted to the container that can be used to persist the python packages. This volume location could be added to the `$PYTHONPATH` environment variable so Ansible picks it up. On startup, the container could read `${SEMAPHORE_CONFIG_PATH}/requirements.txt`, and install the requirements to the volume. Each subsequent startup would then be able to rely on the packages already being installed.

Please let me know if this is something that would benefit the project! I'm happy to open a PR implementing this enhancement if it would be useful.

dmbonsall:
Resolves #1245

Documentation: https://github.com/fiftin/ansible-semaphore-docs/pull/17


dmbonsall:
Hello @fiftin, would you mind reviewing this PR?

mjkl-gh:
Would be interested in this as well

caiodelgadonew:
Hello @fiftin , any plans for releasing this??

snvad:
I join those who are waiting this feature

ansibleguy:
Same issue: https://github.com/ansible-semaphore/semaphore/pull/1498

hackathi:
I also need this feature. Is this blocked by something?

tboerger:
It should add options for Python AND Ansible collections installation.

dmbonsall:
> I also need this feature. Is this blocked by something?

@hackathi, this this has sat a while because @fiftin has not responded to the PR, and they are the only person with write access as far as I know. If we can get bandwidth from them, or someone else with write access to provide feedback, then this can move forward.

dmbonsall:
> It should add options for Python AND Ansible collections installation.

@tboerger, given that Semaphore already supports this through a file placed in the project directory at `collections/requirements.yml`, wouldn't adding a collections install step be redundant to that mechanism? Having two install steps for that would also mean that the runner would have to take care to resolve the dependencies installed at startup plus the dependencies installed for the task.

hackathi:
> > I also need this feature. Is this blocked by something?
> 
> @hackathi, this this has sat a while because @fiftin has not responded to the PR, and they are the only person with write access as far as I know. If we can get bandwidth from them, or someone else with write access to provide feedback, then this can move forward.

Thank you for your insight! Well then let's wait together; if there were some technical blockers I'd have looked into contributing towards those.

lima-faux:
also waiting on this 🙏

m0canu1 (approved):

m0canu1:
me also waiting for this

mjkl-gh (approved):

PR #1259, opened 2023-06-06
fix: mysql connect no keep alive

Fushengliangnian:
No operation for a long time. mysql is `invalid connection`

### DIFF
--- a/db/sql/SqlDb.go
+++ b/db/sql/SqlDb.go
@@ -15,6 +15,7 @@ import (
 	"regexp"
 	"strconv"
 	"strings"
+	"time"
 )
 
 type SqlDb struct {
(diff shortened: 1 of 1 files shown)


fiftin:
Hi @Fushengliangnian 

Thank you for you PR!

I think `Close` should stop pinging database.

I will review it soon.

Fushengliangnian:
Thank recover.

This question has been bothering me for a long time.

I hope you can let me know when the repair is finished.

My English is very poor, please forgive me if the expression is not accurate

Thanks a million!!!

andreas-marschke:
@Fushengliangnian @fiftin I think this isn't the best solution to the problem.
A better one would be to review this: http://go-database-sql.org/connection-pool.html and see if we can allow for the MaxIdle connections to be configurable including timeout so that people can adjust this based on their MySQL server setup.

PR #1281, opened 2023-06-17
Add py3-netaddr library to support copying templates between nodes

seniordevops7:
The main reason for this PR is installing the py3-netaddr python library to support copying templates between two or more nodes. I found this because running a k3s playbook to install a k3s cluster.

image

After installs the python library via container cmd:

image

### DIFF
--- a/deployment/docker/prod/Dockerfile
+++ b/deployment/docker/prod/Dockerfile
@@ -10,7 +10,7 @@ RUN apk add --no-cache -U libc-dev curl nodejs npm git gcc g++ && \
 FROM alpine:3.16 as runner
 LABEL maintainer="Tom Whiston <tom.whiston@gmail.com>"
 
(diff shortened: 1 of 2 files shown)


MarkLFT:
Can someone please confirm if this PR will be included in the build?  And if so, a likely timeframe. I love this project, for my work this has the potential to save me lots of time, but currently, because we use k3s, it is all but unusable.  I am currently trying to decide whether to wait, or abandon my docker installation, and try it on a bare server.  For simplicity, I would prefer to keep it docker based, like all the all the other similar tools and services we run.

dgibbs64:
This is a duplicate of my pr https://github.com/ansible-semaphore/semaphore/pull/1198 which as been open since March. unfortunately currently waiting for it to be merged.

seniordevops7:
I'm closing this one prior #1198

PR #1283, opened 2023-06-19
feat(enhancement): use system dark mode

rmtsrc:
Currently by default semaphore is shown using the light theme and does not change based on the current system dark mode setting.

I often find myself toggling the system light theme during the day and the dark mode at night. Most browsers, websites and my IDE all automatically update to reflect the system setting.

This PR uses the browsers `prefers-color-scheme` media query, which takes into account your system theme settings.

[Screencast from 2023-06-19 19-07-04.webm](https://github.com/ansible-semaphore/semaphore/assets/144435/b6b48cb6-b0ad-4f4b-8f69-30e13f211f08)


fiftin:
Hi @rmtsrc 

It is really cool!

fiftin:
I will test it soon. I think we should add switcher: Dark, Light, System.

PR #1387, opened 2023-08-01
fix: set task status when stopping it

ansibleguy:
Should allow users to manually stop tasks that miss their process (_orphaned_)
For now only manual editing of the database task-table can remove them.

Related to:
* https://github.com/ansible-semaphore/semaphore/issues/1385
* https://github.com/ansible-semaphore/semaphore/issues/1376
* https://github.com/ansible-semaphore/semaphore/discussions/1384
* https://github.com/ansible-semaphore/semaphore/issues/1311

\- AnsibleGuy


ansibleguy:
Note: I think it would be clean to check all active tasks on service-startup and set orphaned ones to failed

PR #1394, opened 2023-08-05
feat: config validation, loading settings from env

ansibleguy:
Greetings!

I've recognized that some issues are related to missing config-validation on service-startup!

This PR adds:
* basic config validation using regex
* overriding settings with environment-variables if they were provided
* setting default-values will be set dynamically
* **tests** for all of the mentioned features
* cleaned up special cases for single settings

This type of config validation has its limits, but is easily understandable and extendable.

I completely understand that this is a major change and needs to be tested thoroughly.

Basically the config would be loaded like this:

```go
func ConfigInit(configPath string) {
       fmt.Println("Loading config")
       loadConfigFile(configPath)
       loadConfigEnvironment()
       loadConfigDefaults()

       fmt.Println("Validating config")
       validateConfig(exitOnConfigError)
```

Related to issues:
* https://github.com/ansible-semaphore/semaphore/issues/1339

### DIFF
--- a/db/AccessKey.go
+++ b/db/AccessKey.go
@@ -198,7 +198,7 @@ func (key *AccessKey) SerializeSecret() error {
 		return fmt.Errorf("invalid access token type")
 	}
 
-	encryptionString := util.Config.GetAccessKeyEncryption()
+	encryptionString := util.Config.AccessKeyEncryption
 
 	if encryptionString == "" {
 		secret := base64.StdEncoding.EncodeToString(plaintext)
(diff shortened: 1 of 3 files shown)


fiftin:
Hi @ansibleguy 

Thank you for your job!

But I don't like this approach. If you want to map env vars to config structure, please use tags like for json. For example:

```go
type ConfigType struct {
  EmailAlert    bool   `json:"email_alert" env:"EMAIL_ALERT" rule:"^:([0-9]{1,5})$"`
}
```

ansibleguy:
Thank you for checking & merging it.
My approach/thought was 'doing some validation is better than none' (;

PR #1286, opened 2023-06-20
ci: update Docker to alpine:3.18 / golang:1.20.5

cedbale:

PR #1470, opened 2023-09-07
Update base docker image from `alpine:3.16` to `alpine:3.18`

rehanone:
This PR is picking up what is seams to have been abandoned in #1286. It looks like the changes in that PR were not complete which this PR should do. Also, in order to minimise the impact of the change, I limited only to the base image version update from `alpine:3.16` to `alpine:3.18` and leave the golang version as is. The home is that this will allow for use of newer version of `ansible` as there are newer features that cause the deployments to fail when using this image.


rehanone:
@fiftin can you please review this and let me know if I have missed anything.

PR #1472, opened 2023-09-09
Config validation

fiftin:


ansibleguy:
Just took the time to read your changes - must say:
I really like your solution of setting regex & env-var-name as tags. 👍🏼 (:

Issue #1484, opened 2023-09-13
Alert Proxy settings


aaronnad:
Currently with the support for alerts via Telegram / Slack and email, some of these use Web Endpoints. Ideally it would be nice to be able to configure these to go via a proxy instead of requiring direct access.

Could a config setting be added for the likes of the slack_url called something like `alert_proxy_endpoint: "",`

This would then be used to proxy the connections for the alert urls (similar to setting the https_proxy) but instead its only for these alerts and not system, nor application wide.

Thanks

aaronnad:
PR raised covers the ability to use a proxy specifically for the alerting URL configurations.

For those environments that are behind firewalls / proxies for public internet access, but don't need to go via proxies for internal access, this PR would suit those security needs

PR #1486, opened 2023-09-14
Add in Proxy URL for Alert

aaronnad:
This PR Adds in the Proxy parameter to to send alerts outbound for #1484

The user is prompted for the Proxy URL when using the setup command. If the user leaves the url blank, then the proxy settings are left as default.

I have only been able to validate this using Slack as that's all I use.

But i don't see a reason why the same code won't work on the Telegram function either. If anyone wants to test and validate the Telegram alerts using the proxy URL settings, then feel free to do so.

You could potentially use different proxies for the different alerts if the user so wished, but I figure most environments that are using proxies will use a standard proxy for the hosts.


aaronnad:
Accidentally closed this one when doing a fork resync.
I've re-opened under 1487 with the current new changes

PR #1487, opened 2023-09-14
Add in Proxy URL for Alert for #1484

aaronnad:
This PR Adds in the Proxy parameter to to send alerts outbound for [#1484](https://github.com/ansible-semaphore/semaphore/issues/1484)

The user is prompted for the Proxy URL when using the setup command. If the user leaves the url blank, then the proxy settings are left as default.

I have only been able to validate this using Slack as that's all I use.

But i don't see a reason why the same code won't work on the Telegram function either. If anyone wants to test and validate the Telegram alerts using the proxy URL settings, then feel free to do so.

You could potentially use different proxies for the different alerts if the user so wished, but I figure most environments that are using proxies will use a standard proxy for the hosts.


aaronnad:
@fiftin would be good to see your thoughts on this. It adds extra security controls for those in environments where proxies are deployed. Those that don't have them needn't configure it.

aaronnad:
@fiftin Can this be re-reviewed please. Would be nice to have this setup so we can proceed with resolving the ability to proxy the notifications.

aaronnad:
Can this be looked into please. Its a small non-impacting change to the working of the built in alert system providing the users the ability to restrict the required access for features to work in secure environments.

aaronnad:
@fiftin any potential to look into this minor change that improves a lot of functionaility in security controlled environments please?

PR #1500, opened 2023-09-19
[WIP]: Handle ssh keys with internal ssh agent

elbandi:
Put ssh key to an internal ssh agent, and use that agent for ssh connections. This will solve the issue if ssh sould be use jumphost.

    Host foo
      Hostname 1.2.3.4
    
    Host bar
      Hostname 5.6.7.8
      JumpHost foo


elbandi:
some code come from https://github.com/sourcegraph/sourcegraph/blob/main/cmd/gitserver/server/sshagent/ssh_agent.go

fiftin:
Hi @Elbandi 
It is must important feature. Thank you!

PeterFalken:
@fiftin , @Elbandi - Would it be possible to set a jumphost from the inventory object, so that we can bypass adding an `.ssh/config` file on the container?

hmoffatt:
Thanks for this, it's great to support ssh keys with a passphrase. But if both the key and the passphrase are in the database is it much more secure?

Is there any way that one or the other could be provided in another way, for example running a script to fetch them from a remote secret store?

I guess it is also possible to run an agent in the environment where Semaphore is started with the same key already added and unlocked?

hmoffatt:
I have noted a bug in this here https://github.com/ansible-semaphore/semaphore/issues/1513 - the agent is killed after a play runs, but ssh persistent connections are still alive which may be forwarding the now invalid agent.

elbandi:
> @fiftin , @Elbandi - Would it be possible to set a jumphost from the inventory object, so that we can bypass adding an `.ssh/config` file on the container?

Our system has more jumphost chain not only one. semaphore -> localjump1 -> centraljump2 -> customerjump3 -> server. With different usernames. thats hard to handle without a good ssh config.

PR #1582, opened 2023-10-21
translate into Brazilian Portuguese

alexporto2200:
translate into Brazilian Portuguese


alexporto2200:
Could anyone tell me if this is correct?
I would really like you to accept my PR!

PR #1584, opened 2023-10-21
Add: Support for Microsoft Team Channel Alerts (via Adaptive Card webhook)

robinmalik:
## Add: Support for Microsoft Team Channel Alerts (via Adaptive Card webhook)

A very simplistic PR to add support for pushing error alerts into Microsoft Team Channels by webhook, based on the existing Slack example.

image

### Notes

- The Adaptive Card template JSON is left intentionally uncompressed as this will make for easier adaptation/extension. Ideally this would go into a dedicated template.json, but having never used GoLang before I didn't want to risk changing any of the internals for this project 😄 
- The 'Task URL' to the task only works if `web_host` config value is set in `config.json`. After following the contribution guide to spin up a local instance on http://localhost:3000, this was actually empty for me. The same applies when following the Docker instructions.


robinmalik:
Unsure why the test integration stage is failing with:
```
Aborting on container exit...
task: Failed to run task "dc:up": exit status 1
Error: Process completed with exit code 201.
```

fiftin:
@robinmalik thank a lot!

fiftin:
And sorry for delay ;)

Issue #904, opened 2022-04-05
Survey Variables


jhaesselbacher:
For Survey Variables when editing a template, it is possible to provide a Type multiple choice list for string or integer ?
Multiple choice options with a single selection  for the variable ?
it is also possible to provide a minimum and maximum number of characters when user entering the string ?
it is also possible to provide for entering a password type variable for which the characters will be masked ?

Twim4:
Would be interested in that too. To have more types of surveys

AndrewSav:
Yep, I would also like to indicate that this is the top feature on my list missing. Static surveys would be an improvement, but dynamic surveys, a-la rundeck would be ideal.

fiftin:
> Yep, I would also like to indicate that this is the top feature on my list missing. Static surveys would be an improvement, but dynamic surveys, a-la rundeck would be ideal.

What is dynamic surveys? I didn't use Rundeck.

AndrewSav:
@fiftin Something like [this](https://docs.rundeck.com/docs/manual/job-options.html#remote-option-values) and [this](https://docs.rundeck.com/docs/manual/job-options.html#option-model-provider) The entire page is actually relevant as a primer what a more advanced survey interface can achieve. Also "Cascading Remote Options" section a bit further down is of a particular interest to me.

In short, you can fetch options and their possible value from remote URLs and you can have selection of options/answers depending on the values entered by the user for other options.

Sohn123:
A static multiple choice option would be really great.

ySp-chld:
This issue seems to have a lot of traction and several people liked to have some better survey variable see #1157
Also i'd add maybe if variable is empty having a default value would be great.

SupremeGit:
Until survey variables are improved,, Semaphore is nothing but a useless annoying toy, and a waste of time for anyone attempting to use if for anything useful.

Survey variables are currently so useless and inflexible, it is a disgrace. You don't even have a boolean type FFS.

Why is this request being ignored?

AndrewSav:
@SupremeGit, first of all no one owes you anything here, if you chose to waste your time attempting to use something you cannot use, this is on you. You are receiving free software with no obligations whatsoever, take it or leave it.

[...]

PR #1598, opened 2023-10-27
Add a new type secret in Survey Variable

sparsick:
Sometimes you want to add a password etc. and hide their value. Therefore, I add a new type 'secret' so this input field will hide the typed value is hidden. 

Relates to #904

### DIFF
--- a/web/src/components/SurveyVars.vue
+++ b/web/src/components/SurveyVars.vue
@@ -122,6 +122,9 @@ export default {
       }, {
         id: 'int',
         name: 'Integer',
+      }, {
+        id: 'secret',
+        name: 'Secret',
       }],
     };
   },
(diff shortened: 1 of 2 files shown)


sparsick:
@fiftin Is this PR okay, or should I do more so that it can be merged?

joseangel-piqueras:
I need this feature! 💯

tboerger (approved):

tothakos-code:
This feature could be misleading. It just hides the input and doesn't handling the actual value as a secret thus someone password will be saved in clear text. The value is easily readable inside the browsers network tab.

My improvement idea:

- The secret survey "key:value" pairs would be sent to the backend in a different key than "environment" thus it will be separated.
- Handling the secrets on the backend only need to differ from environments in one place: not saving them to the db.
- This will break the rerun function, because there will be missing variables, so secret survey fields should always be empty and required for a rerun.

I didn't look into how hard it is to implement this idea or if it's even possible. I might later this week, when i have time.

fiftin:
Merged @tothakos-code solution. Thank you.

PR #1604, opened 2023-11-01
fix: add To mail header

xnilsit:
As described in https://github.com/ansible-semaphore/semaphore/issues/1603, this fixes the missing To header in email alerts that may cause spamfilters to block the email.

To was added to emailTemplate and Alert struct.
Some stuff needed to be moved within the for loop:
mailBuffer is reset in each loop iteration, so that the mail content is not duplicated.
alert.To is reassigned to the email address of the user and the template is re-executed with the new values.


xnilsit:
Closed this and opened https://github.com/ansible-semaphore/semaphore/pull/1636 using a seperate branch, so the pull bot won't break it again.

PR #1656, opened 2023-11-30
Enhance installation using Kubernetes

qyqcswill:


fiftin:
Hi @qyqcswill good instruction, but it is too large for README.md. 

We have repo for docs: https://github.com/semaphoreui/ansible-semaphore-docs

I will be very grateful if you create a PR there.

samerbahri98:
maybe put the manifest in `deployment/k8s/manifest.yaml` and write the instructions to run kubectl apply from GitHub.
example: 
```shell
kubectl apply -f https://raw.githubusercontent.com/ansible-semaphore/semaphore/<TAG>/deployment/k8s/manifest.yaml
```

tboerger:
Maybe it would be better to define Kustomize manifests within the deploy directory and automatically render a single manifest from that? That way you can provide all relevant parts for simple `kubectl apply` and users of Kustomize at the same time. If you don't have the time/engery to handle that I can create a PR for that.

Besides that I am also publishing a Helm chart at https://artifacthub.io/packages/helm/cloudhippie/ansible-semaphore which gets automatically updated with new Semaphore releases.

qyqcswill:
> Maybe it would be better to define Kustomize manifests within the deploy directory and automatically render a single manifest from that? That way you can provide all relevant parts for simple `kubectl apply` and users of Kustomize at the same time. If you don't have the time/engery to handle that I can create a PR for that.
> 
> Besides that I am also publishing a Helm chart at https://artifacthub.io/packages/helm/cloudhippie/ansible-semaphore which gets automatically updated with new Semaphore releases.

It could't be better now! Close it.

tboerger:
> It could't be better now! Close it.

I guess this refers to the published Helm chart? :D

PR #1660, opened 2023-12-02
collections should be defined at the project root and not relative to playbooks path

RusticPotatoes:
collections was being referenced from the playbooks relative directory and should be from the projects root directory.  updated roles for consistency on path joining.  thanks @andreas-marschke for the original find, updating the current codebase as there has been so many changes.

if a playbook is in <proj_root>/sub_folder/playbook.yml the previous code expects collections to be in the sub_folder directory <proj_root>/sub_folder/collections/requirements.yml.  

my proposed changes sets the collections directory to always be in the projects root folder as it does for roles directory currently.   

Now it will look for collections in <proj_root>/collections/requirements.yml as expected. 

this resolves issue https://github.com/ansible-semaphore/semaphore/issues/1144 for No collections requirements file found

### DIFF
--- a/services/tasks/LocalJob.go
+++ b/services/tasks/LocalJob.go
@@ -395,7 +395,7 @@ func (t *LocalJob) getRepoPath() string {
 }
 
 func (t *LocalJob) installRolesRequirements() error {
-	requirementsFilePath := fmt.Sprintf("%s/roles/requirements.yml", t.getRepoPath())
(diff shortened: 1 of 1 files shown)


fiftin:
Hi @RusticPotatoes sorry, I missed your PR.

I merged this PM which do same that yours: https://github.com/semaphoreui/semaphore/pull/2001

PR #1667, opened 2023-12-05
Polish translation

Pshemas:
My attempt at translating into Polish. Definitely needs closer look - it's my first pull request :D .

Will think about better wording for all the terms with "artifact" word. Also as usual due to the way ending of the words change in Polish some tweaks will be needed.

Terms related to Ansible directly are synced with 2 most popular books on the subject in Polish ("Ansible 2 w praktyce" being one of them).


fiftin:
Thank you!

PR #1668, opened 2023-12-06
Update Cron docs link to v3

Stiverton:
Semaphore is using v3 of Cron but the docs link in the UI was not updated to point to the new documentation.  This fixes that.


fiftin:
Thank you!

PR #1672, opened 2023-12-09
Spanish Translation

AndresJ551:
My contribution to the Spanish translation, It will need a review, specially on the UX, since Spanish strings are longer.
Probably some second opinion on some strings, but it's a start.


fiftin:
Thank you!

PR #1675, opened 2023-12-11
fix: #1669

### LINKED ISSUE #1669: ansible_become_user always provided via extra vars, cannot be overwritten
In a playbook, I want to check out a git repository as a certain user. As the `ansible.builtin.git` module does not support owner/group arguments, I do this by providing the `become_user` option instead, such as this:

```yaml
- hosts: all
  tasks:
    - become: true
      become_user: www-data
      ansible.builtin.git:
        repo: https://github.com/example/example
        dest: /var/www/exampleapp.git
```

This issue is not limited to the git module, but here it becomes the most apparent: When I this playbook manually, it works as intended. However, when run via Semaphore, I get the following error message:

```
fatal: [example]: FAILED! => {"changed": false, "msg": "Could not determine branch data - received ", "stderr": "fatal: detected dubious ownership in repository at '/var/www/exampleapp.git'\nTo add an exception for this directory, call:\n\n\tgit config --global --add safe.directory /var/www/exampleapp.git\n", "stderr_lines": ["fatal: detected dubious ownership in repository at '/var/www/exampleapp.git'", "To add an exception for this directory, call:", "", "\tgit config --global --add safe.directory /var/www/exampleapp.git"], "stdout": "", "stdout_lines": []}
```

This is due to the fact that Semaphore always loads the variable `ansible_become_user` via an `--extra-vars` flag, which [has the highest variable precedence](https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_variables.html#understanding-variable-precedence) and even overrides the `become_user` option of the task.

Semaphore even overrides the `ansible_become_user` variable when the become secret does not have a username defined (which is shown as optional in the UI):
image

In this case, the `ansible_become_user` variable is set to the empty string in the extra vars file:
```json
{"ansible_become_password":"correct horse battery staple","ansible_become_user":""}
```

This is rather un-intuitive (and in my opinion undesired) behavior - when the become user is intentionally not defined, Semaphore should not define it as an empty string instead.

s3lph:
This PR fixes #1669 by only writing the `ansible_become_user` variable to the extra vars file if the become secret actually contains a username.

I tested this with the playbook I was having issues with in #1669.

### DIFF
--- a/db/AccessKey.go
+++ b/db/AccessKey.go
@@ -143,7 +143,9 @@ func (key *AccessKey) Install(usage AccessKeyRole, logger lib.Logger) (installat
 		switch key.Type {
 		case AccessKeyLoginPassword:
 			content := make(map[string]string)
-			content["ansible_become_user"] = key.LoginPassword.Login
+			if len(key.LoginPassword.Login) > 0 {
+				content["ansible_become_user"] = key.LoginPassword.Login
+			}
 			content["ansible_become_password"] = key.LoginPassword.Password
 			var bytes []byte
 			bytes, err = json.Marshal(content)


s3lph:
Hm it appears the CI failure is a psql connection error. @ansible-semaphore could someone of you re-run the CI to check whether this is only transient?

fiftin:
Oh, thank you!

Issue #417, opened 2017-08-10
Backup Project Configuration


pbolle:
In jenkins we export our project configuration and push it git. It would be helpful to do the same with semaphore projects.
I stared with a small python program to export all data for a project via api.
The fist problem I found was that private Keys are not exported via api.

What it the best Way to export a project save it to git and imported. 

Regards Philipp

strangeman:
Saving private keys in GIT is very insecure! Jenkins won't allow it too.
Why you can't just backup mysql database and config?

pbolle:
We have many project using the same playbooks. For this project we like to save the best practice as a project (to git is preferred) to create a new project with all playbooks and configuration as a new project. With a mysql backup I can only backup the whole system but not only one project.
The ssh key is not the main problem for us. For this problem I could use a random generated key. The problem is that it is not so easy to export all to the filesystem and create for the exported data a new project. With the api it is possible but a lot of work.

Any idea how to solve the problem best?

Regards Philipp

twhiston:
It would be pretty easy to use ansible to make all the api calls you need to backup and restore the project, but a full project export/import feature might be nice for the future. It will for sure never export private data like keys of vault passwords though so you will always need to do a little configuration for that

fiftin:
I think it should be Copy Project feature or Project Templates.

samerbahri98:
@fiftin I would like to start working on this if that is okay for you. I also need something similar.

samerbahri98:
If everyone is having this issue and wants to experiment with my solution on #1748 feel free to let me know your opinion.

PR #1718, opened 2024-01-28
fix(api), return correct template start_version when using sql

samerbahri98:
I stumbled upon this bug when I was working on #417 
This change definitely does not close the aforementioned issue, but it is a necessary prerequisite.

The gist of this bug is that when when using sql and when requesting the templates list, the response sets `start_version` to null even if the template is of type build. The fix for that is trivial


fiftin:
Thank you!

PR #1719, opened 2024-01-29
Supplementary Chinese translation

ExplodingDragon:
Simplified Chinese (`zh-hans`) does not represent the common Chinese language, this category is broken down to facilitate other maintainers to provide Traditional Chinese (`zh-hant`), Cantonese (`zh-yue`), and Minnan (`zh-min-nan`).


fiftin:
@ExplodingDragon thank you!

PR #1720, opened 2024-01-29
fix(oidc): parse right claims for GitHub authentication

tboerger:
Before this change the authentication via GitHub always errored out within getProfileNameFromEmail because the rune list had a zero length, after my investigation I have been able to pin it down to a lacking email address fetched from the GitHub provider.

Since there are various claims like the username, email and fullname which aren't part of the generated claims I have added a switch to properly read all relevant information from the right claims within the GitHub provider.

This commit is still lacking proper error display on the login page, but this should be added on another commit in general.

### DIFF
--- a/api/login.go
+++ b/api/login.go
@@ -475,13 +475,38 @@ func oidcRedirect(w http.ResponseWriter, r *http.Request) {
 		userInfo, err = _oidc.UserInfo(ctx, oauth2.StaticTokenSource(oauth2Token))
 
 		if err == nil {
-			claims.email = userInfo.Email
-			claims.username = getRandomUsername()
+			switch pid {
+			case "github":
+				type githubClaims struct {
+					Login string `json:"login"`
+					Email string `json:"email"`
+					Name  string `json:"name"`
+				}
 
-			if userInfo.Profile != "" {
-				claims.name = userInfo.Profile
-			} else {
-				claims.name = getRandomProfileName()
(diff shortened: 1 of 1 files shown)


tboerger:
I have resolved the conflicts from the other merged PR.

fiftin:
Hi @tboerger ,
I figured out this problem. Problem not in the code. GitHub do not return email address of the user until he/she don't add them to the public emails.

fiftin:
Discussing of the same issue: https://stackoverflow.com/questions/35373995/github-user-email-is-null-despite-useremail-scope

fiftin:
Looks like additional API call required for GitHub: https://api.github.com/user/emails

tboerger:
But at least for me it was not only missing the email but also the username and fullname, that's why I have started this PR.

But since my email doesn't get populated I guess I got to enable it within github.

fiftin:
Closing the PR because I fixed issue with auth via GitHub (when email inavailable): https://github.com/ansible-semaphore/semaphore/commit/989be6df6860784dc8db9613b3f7c5fa5e7b4868

fiftin:
Thanks for your PR because I get `userInfo.Claims()` from it :)

tboerger:
Great that this had fixed the GitHub authentication, but are these values expected if I register on Semaphore Cloud?

Screenshot from 2024-02-01 10-35-16

PR #1725, opened 2024-01-30
refactor: using OS instead of ioutil which is deprecated

Caixetadev:
This pr removes all ioutil that is deprecated and adds the OS in its place


tboerger (approved):
This really makes sense because of the deprecation.

andreas-marschke:
Are the whitespace changes necessary? The change itself is A-OK It's just very cluttered.

tboerger:
> Are the whitespace changes necessary? The change itself is A-OK It's just very cluttered.

That's simply what a proper IDE is doing. You can hide the whitespace changes if they annoy you by appending `?w=1` to the URL or by clicking on the settings icon within diff view and enabling "Hide whitespace".

tboerger:
Sorry, #1818 have been merged instead of this pr.

PR #1730, opened 2024-01-31
docker-compose localtime

### LINKED ISSUE #1042: Cron at UTC time instead of local time
I have a cron for a task setup as follows (should be 1am local time) and it's running at 6pm local time, which seems to indicate it's running at 1am UTC

image

image

image

Being that the UI is in my local time I would expect the cron to match that (or vice versa to indicate everything is in UTC).

Is there a setting behind the scenes for the time zone that I can set? I don't see anything in the doc or UI.

cbrookins:
Updates docker-compose example to use the local time of the host

Resolves #1042

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,8 @@ services:
     volumes:
       - /path/to/data/home:/etc/semaphore # config.json location
       - /path/to/data/lib:/var/lib/semaphore # database.boltdb location (Not required if using mysql or postgres)
+      - /etc/localtime:/etc/localtime:ro
+      - /etc/timezone:/etc/timezone:ro
 ```
 
 ## Demo


cbrookins:
https://github.com/ansible-semaphore/semaphore/issues/1042

linsomniac:
Did this work for you?  I made these changes back in Aug in our environment and it did not resolve my timezone issue.

tboerger (commented):
I think it should be enouh to mount the localtime:

```
❯ podman run -ti --rm docker.io/semaphoreui/semaphore:latest date
Fri Feb  2 07:16:53 UTC 2024

❯ podman run -ti --rm -e TZ=Europe/Berlin docker.io/semaphoreui/semaphore:latest date
Fri Feb  2 07:17:04 UTC 2024

❯ podman run -ti --rm -v /etc/localtime:/etc/localtime docker.io/semaphoreui/semaphore:latest date
Fri Feb  2 08:17:44 CET 2024
```

tboerger:
@cbrookins I think adding the tzdata package and proividing the TZ environment variables is a much cleaner solution than mounting timezone information from the host system,  I have created a PR at https://github.com/ansible-semaphore/semaphore/pull/1733.

In the end it's up to @fiftin which solution he prefers :)

tboerger:
I think this PR can get closed now :)

PR #1733, opened 2024-02-02
feat(docker): enable TZ env variable for setting timezone

### LINKED ISSUE #1042: Cron at UTC time instead of local time

tboerger:
Supersedes https://github.com/ansible-semaphore/semaphore/pull/1730 and fixes https://github.com/ansible-semaphore/semaphore/issues/1042.

With the installation of the tzdata package it's possible to provide the TZ environment variable which changes the timezone within the container image.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ sudo semaphore user add --admin --name "Your Name" --login your_login --email yo
 ```
 [![Get it from the Snap Store](https://snapcraft.io/static/images/badges/en/snap-store-black.svg)](https://snapcraft.io/semaphore)
 
-### Docker 
+### Docker
 
 https://hub.docker.com/r/semaphoreui/semaphore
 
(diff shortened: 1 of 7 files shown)


fiftin:
@tboerger did you test it?

tboerger:
Not for the runner itself, but the container is already set to the defined timezone, and since the runner/semaphore is just using the "system" time it should work totally fine:

```
❯ podman run -ti --rm docker.io/semaphoreui/semaphore:latest date
Fri Feb  2 07:49:51 UTC 2024

❯ podman run -ti --rm -e TZ=Europe/Berlin localhost/semaphore:latest date
Fri Feb  2 08:49:42 CET 202
```

Issue #1024, opened 2022-09-05
Projects export or backup/restore


colossus-digital:
Is there a way to export (or backup and restore) all projects from a working machine to another ?

thank you very much

Richardswe:
@colossus-digital the only thing i can come up with is to do a "dump" of the database, i know the users and tasks are stored there.

fiftin:
@colossus-digital 
I agree it is useful feature.

colossus-digital:
> @colossus-digital the only thing i can come up with is to do a "dump" of the database, i know the users and tasks are stored there.

i'm trying right now, there are 2 DBs:

```
+--------------------+
| Database           |
+--------------------+
| information_schema |
| semaphore          |
+--------------------+
```

Should i export both or just the semaphore ?

fiftin:
Just semaphore

colossus-digital:
just fyi

db backup (sql):

`mysqldump  -h localhost -P 3306 --protocol=tcp -u semaphore --password --databases semaphore --column-statistics=0 --no-tablespaces > ./semaphore_db_backup.sql`

db restore (sql):

`mysql  -h localhost -P 3306 --protocol=tcp --password -p semaphore < ./semaphore_db_backup.sql`

remkolems:
FWIW I'm running semaphore in a docker(-compose) container (PostgreSQL). 

I'm going to use [Minio ](https://min.io) and https://github.com/offen/docker-volume-backup as a separate backup solution. If I figure the latter one correctly out, including battle testing, I'll report back here.

> just fyi
> 
> db backup (sql):
> 
> `mysqldump -h localhost -P 3306 --protocol=tcp -u semaphore --password --databases semaphore --column-statistics=0 --no-tablespaces > ./semaphore_db_backup.sql`
> 
> db restore (sql):
> 
> `mysql -h localhost -P 3306 --protocol=tcp --password -p semaphore < ./semaphore_db_backup.sql`

See also https://github.com/offen/docker-volume-backup#run-custom-commands-during-the-backup-lifecycle

remkolems:
## **Create a backup of Ansible-Semaphore with (offen) docker-volume-backup on a S3 storage platform (MinIO)** 

- Notice the labels (mandatory for correct functioning of the docker-volume-backup container)

### Example given: docker-compose.yml
```yaml
version: '3.8'
services:
# =============================
# PostGreSQL
# -------------------------------------------------
# https://hub.docker.com/_/postgres
[...]

PR #1748, opened 2024-02-07
Backup and Restore Projects

### LINKED ISSUE #417: Backup Project Configuration

samerbahri98:
**Description**

Closes #417 and #1024
This PR implements the features requested on several occasions.
It does so by adding API endpoints, and buttons to the UI
backup
restore
This change is backwards compatible as it doesn't remove or modify old functionalities

**Changes**
- A backup API endpoint to backup a specific project
- A restore API endpoint to create a new project from a backup
- A backup button that sends an API call to the backup endpoint and then converts the response to a JSON file and saves it.
- A restore button that reads the JSON file, parses it and sends its content to the restore endpoint in order to create a new project.
- The necessary additions to the API schema and to the dredd hooks

**Some Design Decisions**
- It is possible to manually construct a backup by following the suggested schema.
- The matching between objects (foreign keying) is done by matching the names instead of the IDs. Therefore  it is not possible to backup or restore projects that have two inventories with the same name or something like that (I believe that the ability to create to objects in the same project with the same name is a bug, but I don't know if it is wise to fix it or not for backward compatibility reasons.).
- entities restored will always have new IDs.
- credentials backup: only the name and the type is handled, not the content of the secret itself. 
- `view.title` is renamed as `view.name` in the backup (not anywhere else, just here).
- cron is an attribute of template, instead of being its own object.
- translation was done with google translate

Let me know if there are any questions/bugs/requests that I need to address before merging.

### DIFF
--- a/.dredd/hooks/main.go
+++ b/.dredd/hooks/main.go
@@ -1,10 +1,11 @@
 package main
 
 import (
-	"github.com/snikch/goodman/hooks"
-	trans "github.com/snikch/goodman/transaction"
 	"strconv"
 	"strings"
+
+	"github.com/snikch/goodman/hooks"
+	trans "github.com/snikch/goodman/transaction"
 )
 
 const (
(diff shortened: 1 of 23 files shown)


samerbahri98:
Thank you for merging @fiftin :)

steadfasterX:
@samerbahri98 it seems it does export "survey vars" but does not import them unfortunately

tboerger:
Looks like there need to be done fixes for the restored structures.

btzq:
This is a great feature! Is there any update?

samerbahri98:
> This is a great feature! Is there any update?

It is already being shipped in the latest versions. You can try it on https://cloud.semui.co
There seems to be a bug in the restore functionality (for survey vars) but unfortunately I became too busy with work to fix it for now.

Hugobox:
Hi! Great feature! Looking forward to a fix for the project restore survey variables bug. We currently have a script to reformat the exported projects to fix the survey variables, but would obviously prefer to not have to do that!

PR #1755, opened 2024-02-10
ci: change default dialect to bolt

fiftin:


andreas-marschke (approved):

tboerger (approved):

tboerger:
We want to deprecated Bolt, so let's avoid updating this default :)

PR #1784, opened 2024-02-26
feat: replace packr and add basic cache headers

### LINKED ISSUE #814: Browser is not caching js/css files
Semaphore should respond with appropriate cache control headers for sempahore. 

I am using nginx to proxy sempahore. This is the response:

```
curl -I 'https://example.com/js/app.cbdb12f7.js'

HTTP/1.1 200 OK
Server: nginx/1.18.0 (Ubuntu)
Date: Sat, 15 Jan 2022 14:29:17 GMT
Content-Type: application/javascript
Connection: keep-alive

```

If it returns cache control headers then it would load faster on slower internet connections.

### LINKED ISSUE #1780: Semaphore really slow because of serving css file
The semaphore UI is really slow, and web developer tools show it is taking 15 seconds to load /css/app.888d9ffe.css

Edit: looks like a problem within our network, but this does show up a slow point for semaphore

tboerger:
Since packr have been deprecated in favor of go:embed for quite some time now I have replaced it with plain go:embed functionality which makes it more future proof.

The only downside is that go:embed can use relativ paths only, that's why I have moved the web/dist folder into api/public.

Besides that I have refactored the static asset handling which includes basic last modification and cache control headers to improve the cache cache behavior for static assets.

Fixes #1780
Fixes #814


fiftin:
I didn't find any problems in the code. It works like expected.

fiftin (approved):

PR #1803, opened 2024-03-02
Fix Vue warn NewTaskDialog dev mode

NikitaDef:
DevMove

Login --> Click **Task Templates** --> Click Button **New Template**

[Vue warn]: Invalid prop: type check failed for prop "templateId". Expected Number with value NaN, got String with value "new".

image


fiftin:
@NikitaDef thank you!

PR #1809, opened 2024-03-06
Remove Integration Extractors

fiftin:
- feat: remove extractor
- feat: remove extractor from routes
- feat: remove usage of extractors
- feat: remove extractor from openapi docs
- feat: remove extractors from ui
- feat(ui): breadcrubs


tboerger:
Why have you removed that?

andreas-marschke:
@tboerger Ah I see now, it's not that extractors are deleted, just the interconnect between Webhook and the matchers/extractors to values was taken out. So it's not so bad as it shaves off a navigation and logical subdivision that could've been removed in my V1. So this is prescient and not the worst idea. 
@fiftin There's a tiny bug in the webfrontend that prevents the sidebar to come up.

Issue #1373, opened 2023-07-28
How to handle user credentials in Semaphore?


Ye-Min-Tun:
Currently, my ansible project is used by many people. I don't want to share credentials with each other. I tried the two ways for user credentials.
(1) Create Inventory with own password and sudo password stored in Key Store.
    - But, as the key store can be used by everyone in project, it's like sharing the credential (just not seeing the password).
(2) I setup Survey Variables to prompt the credentials everytime they run the task. It seems work and fit with my requirement.
    - But, I just noticed that the Rerun action button under each task can show all the survey variables at that time. So, everyone can see the password.
**So, How do you recommend to prompt the user credentials for each user?**
 I want all the work are done via Semaphore UI. I don't want the team to login to the server, so I refused to use ansible-vault.

Version: 2.8.75

ansibleguy:
I don't think there is a clean way of handling 'user-specific credentials' for now.

PR #1846, opened 2024-03-19
Add secret survey var type, for passwords

tothakos-code:
Adding a new survey var type Secret so you can prompt for passwords and such on a task run.
Secret type vars are passed to ansible as extra vars, but not saved to the database for security reasons and not returned by the API.
Rerunning a task with a secret survey var will always show an empty field.

This solves the issue #1373 and partly #904

This PR is an improved version of this PR: #1598


fiftin:
Thank you @tothakos-code 
Will review soon.

fiftin:
I'm sorry I don't like this solution because it's a bit misleading: If you run playbook in DEBUG mode you will see the secrets.

fiftin:
It is not secrets, it is masked variables. So I don't see reason to add separate field for this.

tothakos-code:
The need for this feature is you can type passwords at the start of playbook runs and you don't need to store it in a vault committed to a git repo. The secret type is meant for variables that need to be masked for security reasons. In ansible currently there is no way to mark a variable as no_log so it will be always printed, with debug mode enabled sadly.

I think the problem is with DEBUG mode then. With debug mode enabled all variables are  printed to the screen even vaulted ones. I don't think we can do anything about this(at this time). Semaphore saves this output in plain text so all vaulted variables(usually password, api keys, etc..) would be visible also to. I think this is encryption at rest and is a critical security vulnerability.

I had an other idea to put these variables in an vault and pass the vault file to ansible as an argument, but it still prints them in debug mode so we can't go around that. But I think we should do everything to keep secrets secrets. 

Maybe limiting debug mode to project owners only by default and project owners can give out rights to this feature?  Encrypting the debug output or even better not saving them at all because of this, only printing them to the user that run them? Putting a warning to debug mode that every variable will be printed in clear text is a minimum. But these ideas to solve the vulnerability, which would solve the printing problem. Let me know what do you think.

fiftin:
Hi @tothakos-code 
I agree with you... Looks like we need allow DEBUG checkbox only for admins.

PR #1859, opened 2024-03-23
feat: remove flags from messengers from config file

fiftin:


tboerger:
Not sure about that... It's more specific if you have the enable flags.

PR #1872, opened 2024-03-27
Fix language not being recognized correctly

ExplodingDragon:
Support for `en_US` type language formats.


fiftin:
But your changes break other languages.

web/src/plugins/i18.js
```diff
@@ -7,7 +7,7 @@ Vue.use(VueI18n);
 let locale = localStorage.getItem('lang');
 
 if (!locale) {
-  locale = navigator.language.split('-')[0];
+  locale = navigator.language.replace('-', '_').toLocaleLowerCase();
```
fiftin, line 10: Looks like "fr", "ru", etc will not work after this change.

ExplodingDragon: When an exact match cannot be made, the country code will be removed and the match will continue.

图片

ExplodingDragon:
When an exact match cannot be made, the country code will be removed and the match will continue. The test doesn't seem to be a problem.

ru

图片

fr

图片


zh-cn

图片

web/src/App.vue
```diff
@@ -670,6 +670,11 @@ const LANGUAGES = {
 function getLangInfo(locale) {
   let res = LANGUAGES[locale];
 
+  // failback short i18n
+  if (!res) {
+    res = LANGUAGES[locale.split('_')[0]];
```
tboerger, line 675: This should catch the case mentioned by @fiftin

tboerger (approved):

PR #1882, opened 2024-03-30
SQLite support

fiftin:


tboerger:
Both of us have realised we need more changes to get SQLite properly working :)

PR #1929, opened 2024-04-12
Get rid of lib package, use non-generic package names

tboerger:
First of all it's bad practice to create "lib" or "util" packages and beside that this makes refactorings much easier if you got a smaller surface to search and replace for.


fiftin (approved):
I like it, thank you :)

andreas-marschke (approved):

PR #1934, opened 2024-04-13
Fix: Author & Version return values in alerts and Slack template

prakx87:
With Build Version Enabled:
image

Without Build Version Enabled:
image

### DIFF
--- a/services/tasks/alert.go
+++ b/services/tasks/alert.go
@@ -382,7 +382,7 @@ func (t *TaskRunner) alertInfos() (string, string) {
 		author = user.Name
 	}
 
-	return version, author
+	return author, version
 }
 
 func (t *TaskRunner) alertColor(kind string) string {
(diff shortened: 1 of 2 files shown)


services/tasks/alert.go
```diff
@@ -382,7 +382,7 @@ func (t *TaskRunner) alertInfos() (string, string) {
 		author = user.Name
 	}
 
-	return version, author
+	return author, version
```
tboerger, line 385: Good catch 🙈

services/tasks/templates/slack.tmpl
```diff
@@ -13,7 +13,14 @@
                     "title": "Author",
                     "value": "{{ .Author }}",
                     "short": true
+                },
```
tboerger: The comma should only be placed if a version is given

prakx87: I have made changes as suggested and tested it.

image

fiftin:
Thank you!

PR #1970, opened 2024-04-24
docker: install ansible-galaxy contents if desired

commx:
This PR adds the ability to install ansible-galaxy collections and roles on Semaphore startup when using the Docker image, so collections and/or roles don't need to be installed every time a task is launched.

The approach is very similar to what can be accomplished with the "pre-commands" in the systemd service as described in the [manual installation section](https://docs.semui.co/administration-guide/installation_manually#ansible-collections-and-roles-1) of the administration guide.

Assuming you have the following `requirements.yml`:

```yaml
---
collections:
  - name: ansible.posix
    version: "1.5.4"

roles:
  - name: geeringguy.node_exporter
    version: "2.1.0"
```

Mount it to the Semaphore container with the docker compose service entry:

```yaml
---
services:
  semaphore:
    image: semaphoreui/semaphore:latest
    environment: {}
    volumes:
      - path/to/requirements.yml:/etc/semaphore/requirements.yml:ro
```

### DIFF
--- a/deployment/docker/common/semaphore-wrapper
+++ b/deployment/docker/common/semaphore-wrapper
@@ -169,5 +169,14 @@ else
     echoerr "No additional python packages to install"
 fi
 
+# Install ansible-galaxy contents
+if [ -f "${SEMAPHORE_CONFIG_PATH}/requirements.yml" ]; then
+    echoerr "Installing ansible-galaxy contents"
+    ansible-galaxy collection install --upgrade -r "${SEMAPHORE_CONFIG_PATH}/requirements.yml"
+    ansible-galaxy role install --force -r "${SEMAPHORE_CONFIG_PATH}/requirements.yml"
+else
+    echoerr "No ansible-galaxy contents to install"
+fi
+
 # run our command
 exec "$@"


tboerger:
Really unsure about this feature. Installing python dependencies totally makes sense because they are required by collections, but collections itself could also be easily installed as part of the jobs within semaphore.

commx:
> Really unsure about this feature. Installing python dependencies totally makes sense because they are required by collections, but collections itself could also be easily installed as part of the jobs within semaphore.

Yes indeed and that's what we currently do. This feature is basically based on a customer scenario where a lot of collections/roles have to be installed for each task invocation. This takes a few minutes before the actual playbook runs, because their internet bandwidth capacity is quite limited. Installing those during semaphore startup can save quite some time and bandwith, especially in situations where galaxy contents are reused often so they don't need to get installed over and over again.

andreas-marschke:
@commx I use semaphore in a similar scenario. We've simply revised the container as a whole, depend on the original and build our dependencies into the container we build:

```dockerfile
FROM semaphoreui/semaphore:latest

USER root

RUN apk update && apk add py3-pip py3-setuptools alpine-sdk make gcc \
         wget python3-dev jq postgresql-libs postgresql-client

RUN apk add --virtual .build-deps gcc musl-dev postgresql-dev

COPY ./requirements.txt /tmp/requirements.txt
RUN pip install -r /tmp/requirements.txt

# More to do requirements from galaxy etc. etc. etc.

WORKDIR /home/semaphore
USER semaphore
```

PR #1988, opened 2024-05-02
fix(config):Correct ENV Variable lookup typo for SshConfig

andreas-marschke:
A typo prevented tmp to be set separately from ssh config path


tboerger (approved):

fiftin:
Thank you!

PR #1994, opened 2024-05-04
feat: integrate sqlite3 support

tboerger:


fiftin:
I started this work too. But faced migration issue too.

fiftin:
I think we need to get rid of current migrations before adding SQLite.

tboerger:
I'm closing it for now as it doesn't work until we change the migration system.

PR #1995, opened 2024-05-04
feat: implement new storage with gorm and storm

tboerger:


fiftin:
Hi @tboerger 

can I help with the PR?

tboerger:
I haven't continued on that yet but I would like to get some feedback if this is some structure you are feeling comfortable with :)

fiftin:
It is too much refactoring for single PR.
I happy to get rid of SQL but it should be iterative. And without changing interface to avoid getting a huge number of merge conflicts.

PR #2004, opened 2024-05-08
feat: add docker support for arm32v6

### LINKED ISSUE #1682: 32Bit ARM support
I'd like to know whether you plan on adding ARM 32 Bit (`armv7`) support. I see you use `golang` as base image which has `armv7` images available.

Why? Because I'd like to use my old RaspberryPi 3 as a task scheduler :)

tboerger:
Some people request support for arm32v6 because they want to run Semaphore on their old Raspberry devices.

Fixes https://github.com/semaphoreui/semaphore/issues/1682

### DIFF
--- a/.github/workflows/beta.yml
+++ b/.github/workflows/beta.yml
@@ -95,7 +95,7 @@ jobs:
           builder: ${{ steps.buildx.outputs.name }}
           context: .
           file: deployment/docker/server/Dockerfile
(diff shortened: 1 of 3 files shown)


fiftin:
Merge conflict

tboerger:
After merging https://github.com/semaphoreui/semaphore/pull/2010 this should be fine.

PR #2005, opened 2024-05-08
feat: add support for ppc64le

### LINKED ISSUE #1658: semaphore on ppc64le
Hello,

I think the project is mega good. 
Unfortunately, it is not available on the IBM Power platform. 

There is not the package go-task for ppc64le 
Is there another way to compile on ppc64le? 

https://github.com/ansible-semaphore/semaphore/issues/566

tboerger:
There have been people who requested support for ppc64le architecture which should be enabled with this commit for the goreleaser releases as well as container releases.

Fixes https://github.com/semaphoreui/semaphore/issues/1658

### DIFF
--- a/.github/workflows/beta.yml
+++ b/.github/workflows/beta.yml
@@ -95,7 +95,7 @@ jobs:
           builder: ${{ steps.buildx.outputs.name }}
           context: .
           file: deployment/docker/server/Dockerfile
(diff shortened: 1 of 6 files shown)


tboerger:
I guess we should revert it... Looks like the virtualization takes ages to finish the build.

PR #2047, opened 2024-05-23
Fix: Alerts for Deploy task shows Build Task ID instead of Build Version

prakx87:
Added condition to get incoming version for Deploy task:

Build: 
image

Alert for Deploy task showing Build Version: 
image

### DIFF
--- a/services/tasks/alert.go
+++ b/services/tasks/alert.go
@@ -364,6 +364,8 @@ func (t *TaskRunner) alertInfos() (string, string) {
 
 	if t.Task.Version != nil {
 		version = *t.Task.Version
+	} else if t.Task.GetIncomingVersion(t.pool.store) != nil {
+		version = "build " + *t.Task.GetIncomingVersion(t.pool.store)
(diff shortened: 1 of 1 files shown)


services/tasks/alert.go
```diff
@@ -364,8 +364,8 @@ func (t *TaskRunner) alertInfos() (string, string) {
 
 	if t.Task.Version != nil {
 		version = *t.Task.Version
-	} else if t.Task.BuildTaskID != nil {
-		version = "build " + strconv.Itoa(*t.Task.BuildTaskID)
+	} else if t.Task.GetIncomingVersion(t.pool.store) != nil {
```
fiftin: `GetIncomingVersion` do database request. Let's call once this method.

prakx87: will make necessary change

prakx87: Fixed

fiftin:
@prakx87 thank you!

PR #2077, opened 2024-06-02
Support TF/Bash

fiftin:
Terraform/OpenTofu and Bash support.


deployment/docker/server/Dockerfile
```diff
@@ -25,8 +25,13 @@ RUN --mount=type=cache,target=/go/pkg \
 
 FROM alpine:3.19
 
+ARG TERRAFORM_VERSION="1.8.2"
```
tboerger: This should be defined as ENV var instead of an ARG to avoid continous rebuilds of all layers. Beside that it's IMHO a bad idea to bundle these tools. It increases the size of the image for everybody, also for people not using Terraform/OpenTofu/Pulimi integration adn just to avoid any kind of trouble with Hashicorp.

deployment/docker/server/Dockerfile
```diff
@@ -38,6 +43,25 @@ bash curl git gnupg mysql-client openssh-client-default python3 python3-dev py3-
     chown -R semaphore:0 /var/lib/semaphore && \
     chown -R semaphore:0 /opt/semaphore
 
+
+RUN curl -O https://releases.hashicorp.com/terraform/${TERRAFORM_VERSION}/terraform_${TERRAFORM_VERSION}_linux_${TARGETARCH}.zip
```
tboerger: The curl, unzip and rm commands should be part of a single RUN to reduce the numer and size of layers.

deployment/docker/server/Dockerfile
fiftin: @tboerger any ideas how to workaround this?

fiftin: create images:
* `semaphore-v*-terraform`
* `semaphore-v*-ansible`
* `semaphore-v*-all`

?

tboerger: Maybe add logic to download the binaries on demand when a task for it gets defined/starts?

deployment/docker/server/Dockerfile
tboerger: If we really want to get this route it should be something like this, move this block to the `build` image block and just copy it to the finaly one in this section:

```
# renovate: datasource=github-releases depName=hashicorp/terraform
ENV TERRAFORM_VERSION=1.8.4

RUN cd /tmp && \
    curl -sSLO terraform.zip https://releases.hashicorp.com/terraform/${TERRAFORM_VERSION}/terraform_${TERRAFORM_VERSION}_linux_${TARGETARCH}.zip && \
  unzip terraform.zip
```

This is how to copy it into the final image, this also avoids copying license files or something like that into the /usr/local/bin folder:

```
COPY --from=builder /tmp/terraform /usr/local/bin
```

Generally if the architecture naming of the binaries differ you could also use something like this switch case: https://github.com/toolhippie/terragrunt/blob/master/Dockerfile#L11-L25

deployment/docker/server/Dockerfile
fiftin: @tboerger yes, agree with you. We need to allow install required version of TF/Pulumi by demand. And I think Ansible version too.

But currently I see only creating multiple images for different configurations:
```
semaphoreui/server:v2.10.1-ansible2.19 (base image)
semaphoreui/server:v2.10.1-ansible2.19-terraform1.8
semaphoreui/server:v2.10.1-ansible2.19-tofu1.7
semaphoreui/server:v2.10.1-ansible2.19-pulumi1.7 (Pulumi size = 400 mb!)
semaphoreui/server:v2.10.1-ansible2.19-tofu1.7-pulumi1.7
semaphoreui/server:v2.10.1-ansible2.19-tofu1.7-terraform1.8
semaphoreui/server:v2.10.1-ansible2.19-terraform1.8-tofu1.7-pulumi1.7
semaphoreui/server:v2.10.1-all
```

tboerger: Different ansible version can get installed via requirements.txt, IMHO you don't want to offer different versions of terraform/opentofu/pulumi, latest stable versions should be fine.

If the files for the providers are downloaded by need via semaphore you could even add an template option which defines and downloads the binaries.

No need to add that to the image at all.

PR #2090, opened 2024-06-09
Introduce thin and full docker images

tboerger:


fiftin:
@tboerger what reason to create one more image with `cargo`? I tried, `cargo` not required for installation of `ansible`, `boto`, etc. Lets remove it from the image and that's it.

fiftin:
If needed we can create separate image based on `semaphoreui/semaphore`.

tboerger:
Cargo is for sure required for cryptography and this is also some dependency of other packages installed within the image. As describe on chat the thin image stays with the same tag as before, just for giving more flexibility to customize the semaphore image there is the customize tag.

The customize tag starts as root user, which makes it possible to install alpine packages via packages.txt and it also starts with base-build and cargo for being able to install nearly any python dependency. Later on in the start process it's using su-exec to drop the privileges for running the semaphore process.

fiftin:
I'm understood why `cargo` needed (for arm). But building already too long. Let's leave thin image and extend it with full image if required. 

I planning to use this way for creating multiple images for `terraform`, `pulumi`, `opentofu`, etc.

tboerger:
I'm currently experimenting with the installed tools a little bit to get a good image size. I would suggest to simply use the Alpine package for Ansible for the slim version and using pip maybe only within the custom tag as this will boost the container size.

In the end the user can choose one of the images depending on how much he wants to customize it.

tboerger:
The current state WITH python-dev reaches 843MB. Without python-dev you are not able to install so many oython deps... Than we could directly drop the requirements.yml entirely from the slim image and preinstall a specific subset of python modules from Alpine packages only.

PR #2257, opened 2024-08-02
 Fix git pull 

frizmanuch:
If repo is already pulled, but new submodules added -> new submodules will not be downloaded


fiftin:
Hi @frizmanuch did you test it?

frizmanuch:
Yes, after adding additional submodule to a git repo its pulling new submodules, without this changes it goes on errors.

fiftin:
Thank you!

Issue #2283, opened 2024-08-15
Feature: Can Inventory ID be specified when creating tasks


gaoyue1989:
### Related to

Web-Backend (APIs), Ansible (task execution)

### Impact

nice to have

### Missing Feature

https://github.com/semaphoreui/semaphore/blob/cde8515fb13cde353c496bec35f00a336947a219/db/Task.go#L52
I saw that the inventory ID has been defined in the database task table, which means we can pass it in from the API, but it has not been processed in the future. I think we can override the inventory in https://github.com/semaphoreui/semaphore/blob/cde8515fb13cde353c496bec35f00a336947a219/services/tasks/TaskRunner.go#L251-L257 to make this configuration effective

### Implementation

https://github.com/semaphoreui/semaphore/blob/cde8515fb13cde353c496bec35f00a336947a219/services/tasks/TaskRunner.go#L251-L257 
This code can be modified to
```
	// get inventory
	if t.Task.InventoryID!=nil{
		t.Inventory, err = t.pool.store.GetInventory(t.Template.ProjectID, *t.Task.InventoryID)
		if err != nil {
			if t.Template.InventoryID != nil {
				t.Inventory, err = t.pool.store.GetInventory(t.Template.ProjectID, *t.Template.InventoryID)
				if err != nil {
					return t.prepareError(err, "Template Inventory not found!")
				}
			}
		}
	}
```

PR #2284, opened 2024-08-15
 add Inventory ID  when creating tasks

gaoyue1989:
see https://github.com/semaphoreui/semaphore/issues/2283

https://github.com/semaphoreui/semaphore/blob/cde8515fb13cde353c496bec35f00a336947a219/db/Task.go#L52
I saw that the inventory ID has been defined in the database task table, which means we can pass it in from the API, but it has not been processed in the future. I think we can override the inventory in https://github.com/semaphoreui/semaphore/blob/cde8515fb13cde353c496bec35f00a336947a219/services/tasks/TaskRunner.go#L251-L257 to make this configuration effective

### DIFF
--- a/services/tasks/TaskRunner.go
+++ b/services/tasks/TaskRunner.go
@@ -249,10 +249,15 @@ func (t *TaskRunner) populateDetails() error {
 	}
 
 	// get inventory
-	if t.Template.InventoryID != nil {
-		t.Inventory, err = t.pool.store.GetInventory(t.Template.ProjectID, *t.Template.InventoryID)
+	if t.Task.InventoryID!=nil{
+		t.Inventory, err = t.pool.store.GetInventory(t.Template.ProjectID, *t.Task.InventoryID)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


fiftin:
Hi @gaoyue1989 will old scenario work if user don't provide inventory id in task?

gaoyue1989:
> Hi @gaoyue1989 will old scenario work if user don't provide inventory id in task?



> Hi @gaoyue1989 will old scenario work if user don't provide inventory id in task?

thinks, now the code if i don't provide inventory id in task  is will be use inventory id in template,and if not find inventory id  is also use  inventory id in template

fiftin:
@gaoyue1989 did you test it?

gaoyue1989:
> @gaoyue1989 did you test it?

企业微信截图_3d546207-ae10-4df8-bc24-a1b92fae4730

I add tow test in askRunner_test.go  one is inventory id in task and  one is inventory id is nil in task   
and I also run the changed docker ues api test

fiftin:
Thank you @gaoyue1989 !

PR #2316, opened 2024-08-29
Add: flag "--force" as option in galaxy install process

FactorT:
Add posibillity to turn off flag "--force" in command `ansible-galaxy install role/collection --force -r requirements.yml`
Permanently flag "--force" may cause some problems with concurrent tasks. Because ansible removes installed roles\collections before new installation.
For example:
```
10:00:00am Task1 runs
10:01:20am Task1 starts galaxy role install process with flag "--force"
10:01:21am Task1 deletes installed roles
10:01:22am Task1 installs roles
10:01:30am Task1 runs playbook and prepares to use a some template from one of the installed roles.
10:02:00am Task2 runs
10:02:20am Task2 starts galaxy role install process with flag "--force"
10:02:21am Task2 deletes installed roles
10:02:22am Task1 tries to use a template from one of roles, but cannot find it and crashes with error
10:02:23am Task2 installs roles
10:02:30am Task2 runs playbook
```
Therefore, the ability to disable the flag "--force" is required to avoid deleting and installing already installed versions of roles/collections


fiftin:
Hi @FactorT 

Perhaps we need separate directory with installed galaxy roles/collections for each template?

PS:
I'm sure adding global flag for all templates is not the best idea. I remember how many requests I got for `--force`.

FactorT:
@fiftin do you mean we need to use flag:
```
-p ROLES_PATH, --roles-path ROLES_PATH
                        The path to the directory containing your roles. The default is the first writable one configured via DEFAULT_ROLES_PATH: {{ ANSIBLE_HOME ~
                        "/roles:/usr/share/ansible/roles:/etc/ansible/roles" }} . This argument may be specified multiple times.
```
for command `ansible-galaxy install`? And install all collections and roles into the corresponding template folder?
For example: /tmp/semaphore/repository_26_10/roles ?

fiftin:
@FactorT yes, I think it can be used.

Issue #2329, opened 2024-09-04
Problem: Opentofu workspaces not supported with backend http


Khaz66:
### Issue

I tried to launch my opentofu code with semaphore UI. In my organization, we use our own git to store the tf.state so we use a backend http.

When i launch the task i have a message :

![image](https://github.com/user-attachments/assets/5221f392-4f8a-41b4-83ae-dd6fcf8ebbb4)

It seems that the backend http doesn't support workspaces. Would it be possible to make the use of a workspace optionnal ?

### Impact

Other

### Installation method

Binary

### Database

MySQL

### Browser

Firefox

### Semaphore Version

2.10.22-e44910d-1721658299

### Additional information

OpenTofu v1.8.1

jriberg:
I have the same issue. Haven't found any workarounds

jriberg:
I made a fix here in this PR. I'm using my branch right now in my own environment.

[Pull Request](https://github.com/semaphoreui/semaphore/pull/2357)

jriberg:
you can tryout my updated docker image from [dockerhub](https://hub.docker.com/r/joafri/semaphoreui)

zancres:
I've been having the same error when trying to use terraform remote state (HCP backend) since it uses a custom workspace I would think.

jriberg:
@zancres please try my image that I linked above, it have my fix in it and works good for me both at home and at the company I'm working for at the moment

salustre:
@fiftin Please tell me are there any chances to fix this bug soon? The change proposed by @jriberg looks reasonable, maybe there could also be an option not to choose workspace at all, in the job definition. Best regards.

fiftin:
> @fiftin Please tell me are there any chances to fix this bug soon? The change proposed by @jriberg looks reasonable, maybe there could also be an option not to choose workspace at all, in the job definition. Best regards.

The PR solves one problem and create another.

The ability to turn off workspaces looks not very good for me.

I think we need to check if workspaces works for a repo.

salustre:
I believe there's still no support for workspaces, when using the http backend.

salustre:
Some workaround should be implemented for sure, because in the current state the opentofu/terraform functionality i pretty much useless for http backend users.

Khaz66:
It is now corrected in the new version thank you.

PR #2357, opened 2024-09-24
Bugfix/2329 workspace and http backend

jriberg:
Improved the logic behind tofu/terraform init
Now the command to chose workspace is not ran if the workspace name is 'default' in the inventory. 
Maybe there is a better way to control this. I think documents need updating too.
Alternatively is to add the ability to not name the workspace so an empty name wont trigger the workspace selector instead of using "default"
                
 Related to https://github.com/semaphoreui/semaphore/issues/2329"


fiftin:
This change can have side effect:
If someone change workspace to Dev for example save Template. If he tried change to workspace back to default, he/she can't do this.

Issue #2325, opened 2024-09-02
Feature: Notifications to other targets, than Telegram/Slack/Email


nuclon:
### Related to

Configuration

### Impact

major improvement to user experience

### Missing Feature

It would be great to have options to send notifications to another targets, not just Telegram/Slack/Email

### Implementation

There is a library, which supports a large number of targets for sending notifications:
https://github.com/nikoksr/notify

shaxiaozz:
Is DingTalk Webhook supported?

nuclon:
Yes, DingTalk is supported by the library, I mentioned

PR #2363, opened 2024-09-26
feat: Add dingtalk webhook notification medium

shaxiaozz:
Hello, we need to use DingTalk's webhook notification, so we added the code of this notification medium, hoping to merge it in. Because we will push the platform to the production environment for formal use. Thank you~

This code has been tested and verified in the local development environment, and the following is the effect diagram:
image


shaxiaozz:
Related issue: https://github.com/semaphoreui/semaphore/issues/2325

fiftin:
Hi @shaxiaozz ,

Thank you for your PR! I removed the block of code which asks dingtalk options from `semaphore setup` because it is outdated method of configuration.

shaxiaozz:
@fiftin Thank you for your agreement to merge. At the same time, delete the obsolete code, thank you~

PR #2387, opened 2024-10-02
Adapt logs for Ansible Galaxy roles and collections

bykof:
adapt logs, so one can better understand, which files are scanned


fiftin:
Thank you!

PR #2392, opened 2024-10-04
feat(be): Allow multiple Ansible vaults

### LINKED ISSUE #1002: Improvement: Support multiple Ansible vault-id instead of just one
Given a ansible.cfg having
`vault_identity_list = dev@.dev-vault-file, database@.database-vault-file`
and according to https://docs.ansible.com/ansible/latest/user_guide/vault.html#passing-multiple-vault-passwords, 
it should be possible to use `ansible-playbook --vault-id dev@.dev-vault-file --vault-id database@.database-vault-file theplaybook.yml` instead of just `ansible-playbook --vault-password-file vaultfile theplaybook.yml`.

As currently implemented, using the 'username' field of Secret's Username&Password as vault identifier 'i.e. dev or database', would allow to create the expected ' --vault-id' for each. 
On the Task's Template, having a multi-select dropdown menu instead of a single entry, would already do it.
Also just selecting one value, should be equivalent to existing behavior:
today:  `ansible-playbook --vault-password-file vaultfile theplaybook.yml`
after: `ansible-playbook --vault-id vaultfile theplaybook.yml`, vaultfile name coming from username from Password form.

https://github.com/ansible-semaphore/semaphore/blob/develop/services/tasks/runner.go#L704

Omicron7:
This adds the ability to specify multiple Ansible vaults during a Task run. 

It changes the UI to allow specifying multiple Access Keys with an associated `name` field. If name isn't specified, the vault is assumed to be the `default`. Only one default vault is allowed. Multiple vaults will be passed to Ansible using `--vault-id={name}@prompt`. I've also added support to backup/restore as well as remote runners.

See https://docs.ansible.com/ansible/2.9/user_guide/vault.html#multiple-vault-passwords

Fixes https://github.com/semaphoreui/semaphore/issues/1002


fiftin:
Wow, really cool feature! Thanks a lot! :) I reviewing it.

PR #2360, opened 2024-09-24
Add Ansible Vault Password Client Script support

Omicron7:
Implements #2359

This adds the ability to use Ansible Vault Password Client Script support for decrypting Ansible Vault. See https://docs.ansible.com/ansible/latest/vault_guide/vault_managing_passwords.html#storing-passwords-in-third-party-tools-with-vault-password-client-scripts

Script name must end in `-client` excluding the extension. Vault Passwords were switched from using `--ask-vault-pass` to `--vault-id=@prompt`. Client Scripts and Passwords were tested locally and are functioning properly.

PR #2397, opened 2024-10-07
feat(be): Add Ansible Vault Password Client Script

### LINKED ISSUE #2359: Feature: Ansible vault password client script
### Related to

Web-Frontend (what users interact with), Web-Backend (APIs), Ansible (task execution)

### Impact

nice to have

### Missing Feature

Add the ability to use an [Ansible vault password client script](https://docs.ansible.com/ansible/latest/vault_guide/vault_managing_passwords.html#storing-passwords-in-third-party-tools-with-vault-password-client-scripts) to decrypt a vault.

This feature is useful for using 3rd party services, database, or server keyring to provide the password necessary to unlock an Ansible vault.

### Implementation

This could be implemented as a new `type` of `AccessKey`, which could subsequently be assigned to a `Template`. During Task execution, if a client script is provided, pass it to `ansible-playbook` using `--vault-id=@{script}` instead of prompting for a password.

### Design

_No response_

Omicron7:
This adds the ability to use Ansible Vault Password Client Script support for decrypting Ansible Vault. See https://docs.ansible.com/ansible/latest/vault_guide/vault_managing_passwords.html#storing-passwords-in-third-party-tools-with-vault-password-client-scripts

Script name must end in `-client` excluding the extension. Vault Passwords were switched from using `--ask-vault-pass` to `--vault-id=@prompt`. Client Scripts and Passwords were tested locally and are functioning properly.

The combination of this Pull Request and #2392 will allow specifying multiple Client Scripts or a combination of Client Scripts and Passwords.

This is the same as #2360 but rebased on `develop`. 

Fixes #2359


fiftin:
Hi @Omicron7 

I don't think that storing script file name in the secret is a right approach.

I think we can extend `TemplateVault` model and add fields `type` (with values `secret` or `script`) and `script_filename` to it.

Omicron7:
> I don't think that storing script file name in the secret is a right approach.

@fiftin Makes sense. When I first implemented this, `TemplateVault` didn't exist yet. I'll work on refactoring this.

fiftin:
@Omicron7 Thanks a lot!

Omicron7:
@fiftin This should be good to go. I moved all of the functionality from `AccessKey` to `TemplateVault`.
Not sure why `Codacy` check is giving me issues again.

fiftin:
Thank you! Will review ASAP. Codacy doesn't like SQL migrations. I don't know how to disable this check.

db/bolt/template.go
```diff
@@ -78,6 +78,12 @@ func (d *BoltDb) GetTemplates(projectID int, filter db.TemplateFilter, params db
 			return
 		}
 
+		templates[i].Vaults, err = d.GetTemplateVaults(projectID, templates[i].ID)
```
fiftin: What reason to get vaults here? It creates a lot of requests to database.

As I understand the vault model required only for creating task.

fiftin (changes requested):

db/bolt/template.go
Omicron7: I believe it's needed here so that elsewhere in the code the `Vaults` are available. Without this, the Backup `json` file is missing the Vaults, as well as the Get Templates API endpoint.

fiftin: One second, will check. Thank you

fiftin: Yes, you are right. I was sure I added this to the backups.

I will fix vault backups.

fiftin: Done https://github.com/semaphoreui/semaphore/commit/2fc7f934305edfcb00eccadfce622f3bc79a9b69. So, I'm sure this fragment of code not required anymore.

niklasweimann:
Hey @fiftin 👋 
Is there a plan when this feature will be released? 🤔 v2.10.35 does not contain this fix, right?

Issue #1357, opened 2023-07-21
Feature: Limit as own Task-Form Field


ansibleguy:
Greetings!

`--limit` is a very commonly used flag that deserves its own form-field as the JSON block is not so 'nice' to work with (_especially on a per-task basis_)

Related to:
* https://github.com/ansible-semaphore/semaphore/discussions/1352
* https://github.com/ansible-semaphore/semaphore/issues/1053
* https://github.com/ansible-semaphore/semaphore/discussions/1089
* https://github.com/ansible-semaphore/semaphore/discussions/848
* https://github.com/ansible-semaphore/semaphore/discussions/709#discussioncomment-2785078

fiftin:
Hi @ansibleguy Thank you!

kyberorg:
Hi @ansibleguy , thanks for raising this issue. I personally, use `--limit` almost every run.

ansibleguy:
Possible Workaround for now: ['ansible_limit' via Survey vars](https://github.com/ansible-semaphore/semaphore/issues/1053#issuecomment-1647323315)

stevmills:
@ansibleguy, concur strongly!

craigarms:
I've noticed that the limit field has been created in the Database and API in this commit : 23c3947e183fec353a46e1f05c601bccf514eaec

So i guess Its just a matter of time before it gets introduced into the GUI

I'd like to have the ability on a template to say that limit is required or not in the same fashion as Survey Variables, this could help stop people from launching a playbook on all hosts at once, 

Pushing this idea further would be to be able to put the limit in the template and at run time any limit passed would be additional.

I work with Netbox source inventories and for example would like the user that is allow to run the playbooks on only the dev env have the dev group "hardcoded" in the limit and any additional limit the user enters is for is run "confort"

PierrickBrun:
I've added the field to the view in this commit: https://github.com/PierrickBrun/semaphore/commit/7a3a8c5ec5a4a3d1c22575614a7241020fa47b24 

It worked without adding anything else since the field is already present in the API.

I made it to my own use so I made the field required but if you want to test it I created a release including the commit. https://github.com/PierrickBrun/semaphore/pkgs/container/semaphore

papierkorp:
Is there a update on when the limit field is coming?

r3nor:
I am also looking forward to this

bmcatherasoo:
[...]

PR #2420, opened 2024-10-14
Feat: Adding limit field to template and Taks, overriding the CLI args from task.

Gnyblast:
Fix for https://github.com/semaphoreui/semaphore/issues/1357 and making override of the CLI args from task run dialog:
- Added limits fields to the template form and runs task dialog
- Created DB fields for this new field to be stored
- Changed Models to support parsing from database or api json
- Overriding implemented for the limit from task to template to enable pre-run changes to take effect
- Added language element for "limit"
- Fixed a JS error that was getting thrown due to item elements was being tried to set while item was undefined
- Made run dialog to display the args defined on template to make it obvious for the end-user to see what's going to be ran
- Made it also override the args by the task to template, so that if you change the args on pre-run it's really a override.


services/tasks/LocalJob.go
```diff
@@ -375,12 +375,13 @@ func (t *LocalJob) getPlaybookArgs(username string, incomingVersion *string) (ar
...
-		t.Log("--limit=" + t.Task.Limit)
-		taskExtraArgs = append(taskExtraArgs, "--limit="+t.Task.Limit)
+	if t.Task.Limit != nil && len(*t.Task.Limit) > 0 {
+		t.Log("--limit=" + *t.Task.Limit)
+		taskExtraArgs = append(taskExtraArgs, "--limit="+*t.Task.Limit)
 	}
 
-	args = append(args, templateExtraArgs...)
+	// override the args from task to template
+	// args = append(args, templateExtraArgs...)
```
fiftin: This code required because task can be triggered automatically via Integration or Schedule. Not only from web interface.

So, template's `hosts_limit` must be used on the backend too if task's `hosts_limit` is empty.

Gnyblast: I implemted a new function called IsCLIArgOverriden that compares the task and template cli args one by one to each other and manages to decide if an template argument is being overriden by a task cli argument. The what it does it, replacing the CLI argument of `templateExtraArgs` with task argument that is targeting to override it. This way task arguments are redundent after the comparison because we are managing to override them to templateExtraArgs variable and use that only. This method should be covering any way of running a template.

Gnyblast:
Not sure why codacy fails though...

fiftin:
Thank you! Reviewing

Gnyblast:
Thanks waiting for the response, we are kind of waiting this for migration.

web/package.json
```diff
@@ -3,7 +3,7 @@
   "version": "0.1.0",
   "private": true,
   "scripts": {
-    "serve": "vue-cli-service serve",
+    "serve": "vue-cli-service serve --port 3001",
```
fiftin: I think this change shouldn't be in the PR ;)

services/tasks/LocalJob_test.go
```diff
@@ -0,0 +1,56 @@
...
+		err     error
+	} = []struct {
+		tmplArg string
+		taskArg string
+		want    bool
+		err     error
+	}{
+		{
+			tmplArg: "--ssh-extra-args=\"-p 3222\"",
+			taskArg: "--ssh-extra-args \"-p 3222\"",
```
fiftin, line 21: Single argument can not contain space which separates key and value. It must be 2 separate arguments. Otherwise `ansible-playbook` doesn't accept this argument (tested).

fiftin (changes requested):

services/tasks/LocalJob_test.go
Gnyblast: I didn't get full what you mean? Do you mean that `--ssh-extra-args` will only work if separator is `=` sign? Like: `-ssh-extra-args="-p 3222"`? and won't work with `--ssh-extra-args "-p 3222"`? If that's the case, how do you prevent this on template CLI args declaration already? And this is not a user mistake if that's the case? On the other hand this piece of code only is to determine if they are passing same argument from task CLI args, so that it can be overriden, I don't really touch CLI args how they look like, regex is just checking either first `=` or first white space and splits it from there, so it can compare the passed option, `--ssh-extra-args` in this case.

web/package.json
Gnyblast: Yeah my bad fixing this: wasn't sure what port vue uses by default, so I created the container exposing 3001 for it and had to add it there.

Gnyblast: I pushed the fix for this.

services/tasks/LocalJob_test.go
fiftin: @Gnyblast yes, it is true. Argument `--ssh-extra-args "-p 3222"` doesn't work in Semaphore.

You can try to call ansible CLI with params `ansible-playbook '--ssh-extra-args "-p 3222"'` and you will got error. But following command will work: `ansible-playbook '--ssh-extra-args' '-p 3222'`

Semaphore runs ansible by command like `exec.Command("ansible-playbook", []string{"--ssh-extra-args \"-p 3222\""})` so it will not work. But following code will work: `exec.Command("ansible-playbook", []string{"--ssh-extra-args", "-p 3222"})`

Gnyblast: as I said before this piece of code here is just to test if the parser regex can split argument and it's value correctly by delimiter which is either `=` sign or a space. I don't do any formatting as you can see in the code below:
```
	for _, taskArg := range taskExtraArgs {
		for tai, tmplArg := range templateExtraArgs {
			ok, err := isCLIArgsOverridden(tmplArg, taskArg)
			if err != nil {
				t.Log(err.Error())
			}

			if ok {
				templateExtraArgs = append(templateExtraArgs[:tai], templateExtraArgs[tai+1:]...)
				break
			}

			if slices.Contains(templateExtraArgs, taskArg) {
				templateExtraArgs = append(templateExtraArgs[:tai], templateExtraArgs[tai+1:]...)
			}

		}
	}
```
first piece with `isCLIArgsOverridden` is the logic split them by delimiter and decide if it's same argument key with different value, the:
1- if yes, that means it's an override and delete the one on `templateExtraArgs` since `taskExtraArgs` will override that specific arguments
2- If not and if there are duplicate, then again delete the on in `templateExtraArgs` since `taskExtraArgs` already have a duplicate of it.

So if user does a mistake on how to pass a CLI argument, which they can do it now with the current code, it's a user mistake and should be fixed by passing it correctly but using `=` sign or escaping correctly. Nothing changes there.

fiftin:
Hi @Gnyblast I reviewing.

db/Task.go
```diff
@@ -22,12 +22,13 @@ type Task struct {
...
-	Limit       string  `db:"hosts_limit" json:"limit"`
-	Secret      string  `db:"-" json:"secret"`
-	Arguments   *string `db:"arguments" json:"arguments"`
-	GitBranch   *string `db:"git_branch" json:"git_branch"`
+	Playbook         string   `db:"playbook" json:"playbook"`
+	Environment      string   `db:"environment" json:"environment"`
+	Limit            *string  `db:"hosts_limit" json:"limit"`
+	Secret           string   `db:"-" json:"secret"`
+	Arguments        *string  `db:"arguments" json:"arguments"`
+	RemovedArguments []string `db:"-" json:"removed_arguments"`
```
fiftin, line 30: How this related to `--limit`?

fiftin: - I can not merge this PR, I don't understand what this PR do. 
- This changes can break current API usage. 
- We have flag `Task.AllowOverrideArgsInTask` but it is ignored in the PR.
- `api-docs.yml` reformatted. What reason? This change not related to the PR.

You need `--limit` right? Lets create separate PR for `--limit` and I will merge it after review.

fiftin (changes requested):

db/Task.go
Gnyblast: > I can not merge this PR, I don't understand what this PR do.

Might be beause of too much gap between each review and forgetting the context.

> We have flag Task.AllowOverrideArgsInTask but it is ignored in the PR.

Can you tell me where exactly I remove that so it's ignored? As far as I see you just use that variable to display the CLI args dialog on new task dialog.

> api-docs.yml reformatted. What reason? This change not related to the PR.

Well that's probably because of my yaml formatter, that could be fixed but doesn't matter anymore.

Just note this PR to somewhere, at some point you will understand the current logic failure of CLI args override and it will be helpful.

PR #1836, opened 2024-03-15
Adding alerts support for Rocket.Chat; closing Issue-#1091

oussjarrousse:
This pull-request, adds the support for alerts for Rocket.Chat https://rocket.chat/ (an open source slack alternative) 
Based on the code available in the code-base that supports Slack. It addresses issue-#1091 and hopefully closes it.

The change add two new `config.json` variables:

```
"rocketchat_alert": true,
"rocketchat_url": "the_url_to_the_rocket_chat_web_hook",
```

The pull-request adds the following configuration to the cli/setup/setup.go:

	askConfirmation("Enable Rocket.Chat alerts?", false, &conf.RocketChatAlert)
	if conf.RocketChatAlert {
		askValue("Rocket.Chat Webhook URL", "", &conf.RocketChatUrl)
	}	

It adds the function `sendRocketChatAlert()` to services/tasks/alert.go which is (almost) identical to the existing `sendSlackAlert()`. `sendRocketChatAlert()` is called from services/tasks/TaskRunner.go `SetStatus()` in a similar way to all other alerts (email, telegram, and slack).

It also adds a template file /templates/rocketchat.tmpl that specifies the format of the json request.content expected by Rocket.Chat web hook.

Testing is still required.

PR #2427, opened 2024-10-16
feat: Add Gotify alerts

### LINKED ISSUE #2383: Feature: Gotify Notification Support
### Related to

Service (scheduled tasks, alerts)

### Impact

nice to have

### Missing Feature

It would be very nice if semaphore had gotify support. Gotify is used in many Home Lab Environments and used by a lot of users.

### Implementation

https://gotify.net/docs/pushmsg

### Design

_No response_

tjleach98:
This pull request adds [Gotify](https://gotify.net) as an alert target. It's common in homelab environments but we also use it with [Proxmox VE](https://www.proxmox.com/en/proxmox-virtual-environment/overview) for alerts. Adding this would reduce the number of places we have to check for alerts. This request should also close #2383.

These changes are very similar to those for Rocket.Chat(#1836) and DingTalk(#2363). It does not include any changes to `setup.go` for `semaphore setup` because I saw they were removed from the DingTalk pull request.

This request adds three new config variables.
```golang
GotifyAlert  bool   `json:"gotify_alert,omitempty" env:"SEMAPHORE_GOTIFY_ALERT"`
GotifyUrl    string `json:"gotify_url,omitempty"   env:"SEMAPHORE_GOTIFY_URL"`
GotifyToken  string `json:"gotify_token,omitempty" env:"SEMAPHORE_GOTIFY_TOKEN"`
```

It also adds a template for the json request (`services/tasks/templates/gotify.tmpl`) and a function `TaskRunner.sendGotifyAlert()` which is called the same as other alerts in `TaskRunner.SetStatus()`.

This has been tested in a local development version of Semaphore UI with our production version of Gotify.

### Gotify Example
Screenshot_20241016_085651

### DIFF
--- a/services/tasks/TaskRunner_logging.go
+++ b/services/tasks/TaskRunner_logging.go
@@ -106,6 +106,7 @@ func (t *TaskRunner) SetStatus(status task_logger.TaskStatus) {
 		t.sendRocketChatAlert()
 		t.sendMicrosoftTeamsAlert()
 		t.sendDingTalkAlert()
+		t.sendGotifyAlert()
 	}
 
 	for _, l := range t.statusListeners {
--- a/services/tasks/alert.go
+++ b/services/tasks/alert.go
@@ -418,6 +418,68 @@ func (t *TaskRunner) sendDingTalkAlert() {
 	}
 }
 
+func (t *TaskRunner) sendGotifyAlert() {
+	if !util.Config.GotifyAlert || !t.alert {
+		return
+	}
+
+	if t.Template.SuppressSuccessAlerts && t.Task.Status == task_logger.TaskSuccessStatus {
+		return
+	}
+
+	body := bytes.NewBufferString("")
+	author, version := t.alertInfos()
+
+	alert := Alert{
+		Name:   t.Template.Name,
+		Author: author,
+		Color:  t.alertColor("gotify"),
+		Task: alertTask{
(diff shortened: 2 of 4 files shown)


fiftin:
Hi @tjleach98 did you test it?

tjleach98:
Yes! I setup the development version of Semaphore UI and used the built in tasks for testing. It was setup to alert using our production version of Gotify. The screenshot included in my request message is the notification on Gotify after running the 'Ping Site' task.

JcMinarro:
How can I add Gotify config to my Semaphore Docker Instance??
I tried adding the following environment variable but it seems it doesn't work:
```
SEMAPHORE_GOTIFY_ALERT: True
SEMAPHORE_GOTIFY_URL: 10.0.10.5
SEMAPHORE_GOTIFY_TOKEN: ABe7tVJsVEWWiKD
```

tjleach98:
I believe you need to provide the protocol to the `SEMAPHORE_GOTIFY_URL` variable.

In your case, `10.0.10.5` would become `http://10.0.10.5` or `https://10.0.10.5`. Which one depends on your setup.

Do you see any errors at the end of the log when running a play? Did you enable alerts for the project? You can check at 'Dashboard' -> 'Settings' -> 'Allow alerts for this project'.

JcMinarro:
After I enabled alerts on the settings section and added the HTTP schema to the URL, it started to work, thanks

PR #2430, opened 2024-10-16
Allow Parallel Tasks if GitUrl is path.

jdhayes:
Check if Git starts with '/' (is path not url) then skip limiting tasks. This way we can allow concurrent tasks within the same template , as long as it is safe (ie. not running git pull).


jdhayes:
@fiftin This is very useful, it will allow a Task Template to run multiple tasks without compromising the 1 task per limit for "normal" GitUrl based repositories.

fiftin:
@jdhayes did you test it?

jdhayes:
Sorry for the lack of response.
Yes, I have tested the proposed patch, and I am still currently using this today.

jdhayes:
@fiftin Does the flag `allow_parallel` you mentioned allow parallelism if tasks are under the same project?
Or is that to address something else?

fiftin:
Hi @jdhayes I think this implicit behaviour for local repos might confuse users. We should add a flag in the template settings UI to control this feature.

jdhayes:
OK, either way.
I am not picky on how the feature is enabling parallel tasks, just as long as multiple tasks can run under the same project task template :)

jdhayes:
Any timeline for such a feature?

PR #2438, opened 2024-10-17
Add Ukrainian language support

click0:
Add Ukrainian language support.


fiftin:
Hi @click0 
Thanks a lot for your contribution.

We use GPT-4mini to translate strings to all languages. Out script takes whole `en.js` and translate it to all languages.

What if I add Ukrainian to list of supported languages and you check result of GPT-4?

click0:
> What if I add Ukrainian to list of supported languages 

In my experience, the 1/5-1/10 phrase after AI requires correction.

> and you check result of GPT-4?

Only if he will translate only new phrases or changes in existing ones.
In general - yes, I am ready to check.

PR #2463, opened 2024-10-21
Relative Task-Log width

lug-gh:
The task logs in the semaphore are limited to 1000px width. This means that they are quite narrow and difficult to read on monitors with a high resolution. This value should be relative to the resolution, in my example 90%.
This makes the task logs much easier to read.

### DIFF
--- a/web/src/App.vue
+++ b/web/src/App.vue
@@ -72,7 +72,7 @@
     <EditDialog
       v-model="taskLogDialog"
       save-button-text="Delete"
-      :max-width="90%"
+      :max-width="'90%'"
       :hide-buttons="true"
       @close="onTaskLogDialogClosed()"
     >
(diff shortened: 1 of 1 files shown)


fiftin:
Ability to expand task log window added.

lug-gh:
Hi @fiftin 

First of all, thank you for implementing such an option.

However, I think it is a pity that you closed the PR, as your customization does not necessarily match this PR.

When you open a task, everything around the task window is completely useless anyway, as you can't click on anything, and the things you see have no information content, right?

So with a wider task window, you have the advantage of saving an extra click to make the task window readable. I'm not saying that the fullscreen window is unnecessary or should be removed - on the contrary - there are certainly good use cases for it.
But for easy, and above all legible, viewing of the task windows, a dynamic width of the window would be a great advantage. From a UI/UX point of view, it is currently really disadvantageous.

I would be very grateful if you could reconsider this or explain your point of view :)

lug-gh:
Example: 
image

All these line breaks make it difficult to read, while the free space around it cannot be used.

Here's what it looks like at 90%:
image

PR #2470, opened 2024-10-22
Add config option  to provide environment variables for run apps

fiftin:
In v2.10.32 OS level environment variables has been removed for security reason.

To provide custom environment variables new config option should be used:
* environment var (for Docker):  `SEMAPHORE_ENVIRONMENT_VARS`:
   ```
   -e 'SEMAPHORE_ENV_VARS={"SOME_VAR": "test"}'
   -e 'SEMAPHORE_FORWARDED_ENV_VARS=["VAR1", "VAR2"]'
   ```
* config option `environment_vars`:
   ```
   {
      "env_vars": {"SOME_VAR": "test"},
      "forwarded_env_vars": ["VAR1", "VAR2"],
   }
   ```

Value must be string.

### DIFF
--- a/db_lib/AnsiblePlaybook.go
+++ b/db_lib/AnsiblePlaybook.go
@@ -22,8 +22,7 @@ func (p AnsiblePlaybook) makeCmd(command string, args []string, environmentVars 
 	cmd := exec.Command(command, args...) //nolint: gas
 	cmd.Dir = p.GetFullPath()
 
-	cmd.Env = []string{}
-
+	cmd.Env = getEnvironmentVars()
 	cmd.Env = append(cmd.Env, fmt.Sprintf("HOME=%s", util.Config.TmpPath))
 	cmd.Env = append(cmd.Env, fmt.Sprintf("PWD=%s", cmd.Dir))
 	cmd.Env = append(cmd.Env, "PYTHONUNBUFFERED=1")
--- a/db_lib/AnsiblePlaybook.go
+++ b/db_lib/AnsiblePlaybook.go
@@ -22,11 +22,12 @@ func (p AnsiblePlaybook) makeCmd(command string, args []string, environmentVars 
 	cmd := exec.Command(command, args...) //nolint: gas
 	cmd.Dir = p.GetFullPath()
 
-	cmd.Env = getEnvironmentVars()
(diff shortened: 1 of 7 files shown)


Omicron7:
@fiftin Instead of passing in key/value pairs, could we instead just pass in keys and allow Semaphore to lookup the current value from the OS ([`os.Getenv`](https://pkg.go.dev/os#Getenv))?

```
-e 'SEMAPHORE_ENVIRONMENT_VARS=["SOME_VAR", "ANOTHER_VAR"]'
```

I use AWS ECS to run Semaphore, and ENV parameters are defined in AWS Parameter Store and injected into the docker container at runtime. I'd need to write a script that executes before semaphore launches that reads all the necessary ENV key/values and generates the `SEMAPHORE_ENVIRONMENT_VARS` at runtime. This complicates the hosting/runtime and provides an additional layer where ENV secrets may be exposed. By passing in just the ENV name, this would be drastically simplified.

In addition, I think we should have a list of default ENV that are always included, like `PATH`, `HOME`, and `LANG`.

pascaliske:
I would also prefer it the way @Omicron7 describes it - even a comma separated string of ENV keys would be sufficient in my opinion:

```
-e 'SEMAPHORE_ENVIRONMENT_VARS=SOME_VAR,ANOTHER_VAR'
```

pascaliske:
I would also prefer it the way @Omicron7 describes it - even a comma separated string of ENV keys would be sufficient in my opinion:

```
-e 'SEMAPHORE_ENVIRONMENT_VARS=SOME_VAR,ANOTHER_VAR'
```

fiftin:
@Omicron7  added `SEMAPHORE_FORWARDED_ENV_VARS`. It must be valid json array.

@pascaliske currently it supports only JSON because it is our current realization of de-serialization. Comma separated values will be implemented later.

bence42:
@fiftin A documentation update would have been a minimum. This, together with the removal of os level env vars in v2.10.32 is a breaking change.
I'm sure you're aware that the UI will hand with no information unless `ANSIBLE_HOST_KEY_CHECKING=False` is passed, and now users have to dig thru release notes to learn about SEMAPHORE_FORWARDED_ENV_VARS.

Please be more kind to the users.

Issue #2474, opened 2024-10-22
Problem: user `1001` is not able to use `apk` for `packages.txt`


Sergio-IME:
### Issue

This user
https://github.com/semaphoreui/semaphore/blob/74678c63b3573d4dd396d4405f7ec3a4f85d9b9f/deployment/docker/server/Dockerfile#L92
is not privileged enough to do this
https://github.com/semaphoreui/semaphore/blob/74678c63b3573d4dd396d4405f7ec3a4f85d9b9f/deployment/docker/server/server-wrapper#L201-L207

adding
```yaml
user: root
```
to `compose.yml` works, but that defeats the purpose of a dedicated user

### Impact

Docker

### Installation method

Docker

### Semaphore Version

2.10.32

### Logs & errors

```console
semaphore-1  | Installing additional system dependencies
semaphore-1  | ERROR: Unable to lock database: Permission denied
semaphore-1  | ERROR: Failed to open apk database: Permission denied
semaphore-1  | .
semaphore-1 exited with code 99
```

### Configuration

`compose.yml`
```yaml
services:
  semaphore:
    image: semaphoreui/semaphore:latest
    # works with this
    #user: root
    volumes:
      - ./packages.txt:/etc/semaphore/packages.txt
```

`packages.txt`
```text
bash
```

fiftin:
Oh, yes. We need to remove this.

tboerger:
Initially I have added another socket image which was started as root and dropped the privileges later when starting semaphore to make it possible to customize the bundled packages without the need to build a custom image.

There are python libraries that need   some additional packages.

PR #2476, opened 2024-10-22
ci: remove packages.txt https://github.com/semaphoreui/semaphore/issues/2474

fiftin:
Fix issue with semaphore user's permissions: https://github.com/semaphoreui/semaphore/issues/2474


tboerger:
IMHO this was a useful feature, it just requires running it as root user and dropping the privileges later.

PR #2553, opened 2024-11-20
chore: Improve API docs and fix REST issues.

Omicron7:
This Pull Request came about because I was attempting to use the API docs (`api-docs.yml`) to generate a golang client using [go-swagger](https://goswagger.io/go-swagger/) to interact with the API with the goal of creating a Terraform Provider to manage SemaphoreUI using Terraform.

Most of the changes here are just updating the API docs (and dredd tests) to reflect what the API is already doing. In the process of implementing the Terraform Provider I found that some API Create (POST) endpoints didn't return any response. This works for the Semaphore Web Console where a User can visually see what they just created, but not for an application entirely dependent on the API. I added basic responses, and in the case of sensitive data, made sure it wasn't rendered so that every POST now returns at least the ID of the created resource.

I've used this PR locally to develop an initial release of [terraform-provider-semaphoreui](https://registry.terraform.io/providers/CruGlobal/semaphoreui/latest). It includes most of the project resources needed to create and manage a template. The template resource is missing some features. I plan to add data sources and remaining features as time allows, but wanted to get this PR out there so that others could use Terraform to manage SemaphoreUI.

### DIFF
--- a/db/sql/project.go
+++ b/db/sql/project.go
@@ -11,8 +11,8 @@ func (d *SqlDb) CreateProject(project db.Project) (newProject db.Project, err er
 
 	insertId, err := d.insert(
 		"id",
-		"insert into project(name, created, type) values (?, ?, ?)",
-		project.Name, project.Created, project.Type)
+		"insert into project(name, created, type, alert, alert_chat, max_parallel_tasks) values (?, ?, ?, ?, ?, ?)",
+		project.Name, project.Created, project.Type, project.Alert, project.AlertChat, project.MaxParallelTasks)
 
 	if err != nil {
 		return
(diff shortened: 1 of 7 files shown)


fiftin:
Hi @Omicron7 thank you for your contribution. 

It will take a time to review.

db/sql/project.go
```diff
@@ -110,11 +110,12 @@ func (d *SqlDb) DeleteProject(projectID int) error {
 
 func (d *SqlDb) UpdateProject(project db.Project) error {
 	_, err := d.exec(
-		"update project set name=?, alert=?, alert_chat=?, max_parallel_tasks=? where id=?",
+		"update project set name=?, alert=?, alert_chat=?, max_parallel_tasks=?, type=? where id=?",
```
fiftin: Changing the project type may have side effects. Project type currently used in Cloud only but it is reserved for other purposes too.

There is no possibility to change the project type after its creation.

PR #2577, opened 2024-11-25
Update Dockerfile for Kerberos authentifcation windows

cricrack76:
For use this module : gssapi
krb5
kerberos

https://github.com/semaphoreui/semaphore/discussions/2572

I can use kerberos windows authentification if i create this dockerfile
--
FROM semaphoreui/semaphore:v2.10.35
USER root
RUN apk add --no-cache -U --upgrade libressl-dev cyrus-sasl-gssapiv2 krb5 krb5-libs krb5-dev gcc musl-dev python3-dev
COPY krb5.conf /etc/krb5.conf

### DIFF
--- a/deployment/docker/server/Dockerfile
+++ b/deployment/docker/server/Dockerfile
@@ -79,12 +79,12 @@ RUN chown -R semaphore:0 /usr/local/bin/server-wrapper && \
 
 WORKDIR /home/semaphore
 
-RUN apk add --no-cache -U python3-dev build-base openssl-dev libffi-dev cargo && \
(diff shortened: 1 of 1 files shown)


cricrack76:
Hi,
Doing some tests in docker, the problem seems to come from openss-dev
this is not a package that I added.
I had: libressl-dev cyrus-sasl-gssapiv2 krb5 krb5-libs krb5-dev gcc musl-dev
My test :
image

I have no error but if i tried with openssl-dev :
image
image

cricrack76:
Error statement code

Issue #2606, opened 2024-12-03
Feature: Terraform specify --target


markush81:
### Related to

Other

### Impact

major improvement to user experience

### Missing Feature

Terraforms argument `--target` would be a major improvement in usage. Cause sometimes you want only parts to be done really automatically via schedules

```

  -target=resource    Limit the planning operation to only the given module,
                      resource, or resource instance and all of its
                      dependencies. You can use this option multiple times to
                      include more than one object. This is for exceptional
                      use only.
```

### Implementation

I assume either dedicated UI param for `--target` or implemented via possibility of `CLI args` are used could be a solution.

![image](https://github.com/user-attachments/assets/d32b628a-3f6a-46b5-8977-804788736fee)

```
	var templateExtraArgs []string
	if t.Template.Arguments != nil {
		err = json.Unmarshal([]byte(*t.Template.Arguments), &templateExtraArgs)
		if err != nil {
			t.Log("Invalid format of the template extra arguments, must be valid JSON")
			return
		}
	}

	var taskExtraArgs []string
	if t.Template.AllowOverrideArgsInTask && t.Task.Arguments != nil {
		err = json.Unmarshal([]byte(*t.Task.Arguments), &taskExtraArgs)
		if err != nil {
			t.Log("Invalid format of the TaskRunner extra arguments, must be valid JSON")
			return
		}
	}

	args = append(args, templateExtraArgs...)
	args = append(args, taskExtraArgs...)

```

like in `getPlaybookArgs`

markush81:
I just tried with

![image](https://github.com/user-attachments/assets/df74ea91-8e2b-4dce-ba5d-e4a425f58413)


and

```


diff --git a/services/tasks/LocalJob.go b/services/tasks/LocalJob.go
index d2e0491a..a5c869ea 100644
--- a/services/tasks/LocalJob.go
+++ b/services/tasks/LocalJob.go
@@ -256,13 +256,37 @@ func (t *LocalJob) getTerraformArgs(username string, incomingVersion *string) (a
        }
 
        for _, secret := range t.Environment.Secrets {
-               fmt.Println(secret)
                if secret.Type != db.EnvironmentSecretVar {
                        continue
                }
                args = append(args, "-var", fmt.Sprintf("%s=%s", secret.Name, secret.Secret))
        }
 
+       fmt.Println(t.Template.Arguments)
+       var templateExtraArgs []string
+       if t.Template.Arguments != nil {
[...]

PR #2628, opened 2024-12-14
terraform: add templateExtraArgs and taskExtraArgs

markush81:
to support additional parameters for terraform like

```
-target=resource    Limit the planning operation to only the given module,
                      resource, or resource instance and all of its
                      dependencies. You can use this option multiple times to
                      include more than one object. This is for exceptional
                      use only.
```

https://github.com/semaphoreui/semaphore/issues/2606


fiftin:
Refactored code https://github.com/semaphoreui/semaphore/pull/2689

Issue #2645, opened 2024-12-18
Problem: SemaphoreUI Runner do not remove inventory_* dirs


bernd-mueller-el:
### Issue

Using a runner for tasks will result in this message
```
level=error msg="remove /tmp/semaphore/inventory_9: directory not empty"
```
I think this line have to be like the following
https://github.com/semaphoreui/semaphore/blob/d11b166214883de4431fbf53ac421bb6745d66b3/services/tasks/LocalJob_inventory.go#L92

```
	if err := os.RemoveAll(fullPath); err != nil {
```

Is this correct?

### Impact

Semaphore Project

### Installation method

Package

### Database

Postgres

### Semaphore Version

v2.10.43, v2.11.2

```
# semaphore version
2.10.43-e43af6b-1732277822
2.11.2-0e9490c-17352..
```

### Logs & errors

```
level=error msg="remove /tmp/semaphore/inventory_9: directory not empty"
```

FrontiniFederico:
Did you perhaps find a way to move past this issue? I have the exact same problem with a dockerized semaphore installation.

bernd-mueller-el:
Delete manually. This could be automated via e.g. cron. I had no time until now to build the binary with the change above and test this.

bernd-mueller-el:
Created pull request
https://github.com/semaphoreui/semaphore/pull/2698

fiftin:
@bernd-mueller-el the bug in subject relates with other bug: cloning repository for each task.

openstreem:
Same issue with Linux binary. Watching bug.

kalfeher:
An effective work around for the time being is to reboot.
So long as the directories are created under `/tmp` rebooting the host should clear out the left behind inventory directories. 

As OP stated, `os.Remove(fullpath)` will fail if the directory is not empty. `os.RemoveAll(fullpath)` should work.

fiftin:
Fixed in 2.14.0-beta3

PR #2698, opened 2025-01-15
runner: remove inventory directory when not empty

bernd-mueller-el:
reference this problem issue https://github.com/semaphoreui/semaphore/issues/2645

my local test via

```
task build
task release:test
```

and installing rpm in my testing environment was successfull, inventory directory was removed and the error message in issue is not present.


bernd-mueller-el:
https://github.com/semaphoreui/semaphore/issues/2645#issuecomment-2815454618

PR #2701, opened 2025-01-18
Fixed missing venv in Docker Image, Install Ansible deps as mentioned in your docs

joe-speedboat:
Hi
I added some minior fixes for Docker images.

You can test with my setup as described in:
[https://github.com/joe-speedboat/kube.semaphore](https://github.com/joe-speedboat/kube.semaphore)
on k3s
Or just let met test it.

Thanks
Chris

BTW: I work with AWX most of my time and even if semaphore is missing lot of features from awx, its a great project.
         If you like, I will help you get the most needed features from mid-range enterprise perspective, but I am not willing to buy pro licence for my non comercial home lab and testing.

---
Added ANSIBLE_VENV_PATH variable as exist in Dockerfile
Switched into venv before installing python dependencies 
Install ansible dependencies with venv same way as python

### DIFF
--- a/deployment/docker/server/server-wrapper
+++ b/deployment/docker/server/server-wrapper
@@ -35,6 +35,7 @@ file_env() {
 }
 
 export SEMAPHORE_CONFIG_PATH="${SEMAPHORE_CONFIG_PATH:-/etc/semaphore}"
+export ANSIBLE_VENV_PATH="/opt/semaphore/apps/ansible/${ANSIBLE_VERSION}/venv"
 export SEMAPHORE_DB_PATH="${SEMAPHORE_DB_PATH:-/var/lib/semaphore}"
 export SEMAPHORE_DB_PORT="${SEMAPHORE_DB_PORT:-}"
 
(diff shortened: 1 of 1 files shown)


fiftin:
> If you like, I will help you get the most needed features from mid-range enterprise perspective

Hi @joe-speedboat,
Yes, that would be great!

joe-speedboat:
okay, fine, I'm an ex redhat trainer and consultant and ansible admin of first day, mostly I worked with rundeck and awx/tower.
So I met lot of enterprise requirements and solutions.
- AWX is now IBM, which is moving from opensource to cash cow. (removing auth framework, ...)
- Rundeck is opensource. but ansible integration is ugly and it's not comunity driven
So I am always looking for alternatives to start ansible jobs with nice ui, as well some of my customers do.
The comparison table is not  true btw, never do comparison with a product, you did not use for at least 1 year in production env.

okay, so i will start setting up semaphore in my kubernetes env and this is my first PR, would you please merge?
then i will move my workload from awx to semaphore, we will see.

would you mind if i open up freature requests for things i see as a chance or missing requirement?
and its always up to you what to implement or not ... as long as you have good arguments.

Thanks for doing OpenSource!
Chris

PR #2725, opened 2025-01-30
Add multiple semver tags to docker releases

RoboMagus:
Change to GH Workflow that publishes Docker images.

This PR will add additional tags to each release ([Rationale](https://medium.com/@mccode/using-semantic-versioning-for-docker-image-tags-dfde8be06699)). E.g. when releasing `v2.13.1`, additional tags `v2` and `v2.13` are added.  These allow users to pin their preferred main and/or minor version whilst still pulling in patch releases when they are released.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -87,6 +87,8 @@ jobs:
             maintainer=Semaphore UI <support@semaphoreui.com>
           tags: |
             type=raw,value=${{ github.ref_name }}
+            type=semver,pattern=v{{major}}.{{minor}}
+            type=semver,pattern=v{{major}}
           flavor: |
             latest=true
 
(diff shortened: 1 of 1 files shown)


fiftin:
Looks good. But let's remove `type=semver,pattern=v{{major}}` for now. The major version unchanged since 2016. And I have no plan to change it :)

RoboMagus:
> But let's remove type=semver,pattern=v{{major}} for now. The major version unchanged since 2016. And I have no plan to change it :)

Yeah makes sense, but that is sort of the point though. Should the time arise that the major version does get a bump because of a big breaking change having the option to stay up to date with minor versions automatically but hold of on an update that causes said breaking change. 
Such issues arose recently with e.g. Zigbee2MQTT that upgraded to `v2.x` after about 6 years of minor releases on `v1.x`

I'll be happy to remove the major version tag if you're certain this is not something you want, but as a precaution (i.e. just anything between sticking to `latest` and manually updating for every minor tag) it has a lot of added value.

PR #2736, opened 2025-02-06
fix: Commit hash and overlfow

Gnyblast:
- Commit hash were locking the re-runs to stay attached even if you click x button to close and detach it. I fixed this behaviour to make it detach from the commit and fetch the latest when alet is closed
- The new task dialog title was being overflowed hidden when it was too long. I made new task dialog to grow up to 700px depending on title length

### DIFF
--- a/web/src/components/NewTaskDialog.vue
+++ b/web/src/components/NewTaskDialog.vue
@@ -3,6 +3,7 @@
     v-model="dialog"
     :save-button-text="$t(TEMPLATE_TYPE_ACTION_TITLES[templateType])"
     :title="$t('newTask')"
+    :max-width="dialogWidth"
     @save="closeDialog"
     @close="closeDialog"
   >
@@ -51,9 +52,15 @@ export default {
   data() {
     return {
       dialog: false,
+      dialogWidth: 400,
       TEMPLATE_TYPE_ACTION_TITLES,
       TEMPLATE_TYPE_ICONS,
     };
+  },
+  mounted() {
+    this.dialogWidth = (window.innerWidth > 450 + this.templateAlias.length
+      && 450 + this.templateAlias.length < 700)
+      ? 450 + this.templateAlias.length : 400;
   },
   watch: {
     async dialog(val) {
--- a/web/src/components/NewTaskDialog.vue
+++ b/web/src/components/NewTaskDialog.vue
@@ -1,7 +1,12 @@
 <template>
-  <EditDialog v-model="dialog" :save-button-text="$t(TEMPLATE_TYPE_ACTION_TITLES[templateType])" :title="$t('newTask')"
-    @save="closeDialog" @close="closeDialog">
-    <template v-slot:title={ }>
+  <EditDialog
+    v-model="dialog"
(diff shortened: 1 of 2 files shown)


web/src/components/NewTaskDialog.vue
```diff
@@ -51,10 +52,16 @@ export default {
   data() {
     return {
       dialog: false,
+      dialogWidth: 400,
       TEMPLATE_TYPE_ACTION_TITLES,
       TEMPLATE_TYPE_ICONS,
     };
   },
+  mounted() {
+    this.dialogWidth = (window.innerWidth > 450 + this.templateAlias.length
```
fiftin: What this code do?

Gnyblast:
The new task dialog title was being overflowed hidden when it was too long. I made new task dialog to grow up to 700px depending on title length

web/src/components/NewTaskDialog.vue
Gnyblast: The new task dialog title was being overflowed hidden when it was too long. I made new task dialog to grow up to 700px depending on title length

fiftin: Sorry, I mean this condition `window.innerWidth > 450 + this.templateAlias.length` ?

Gnyblast: it's not a single line, the full code is:
```
this.dialogWidth = (window.innerWidth > 450 + this.templateAlias.length
      && 450 + this.templateAlias.length < 700)
      ? 450 + this.templateAlias.length : 40
```
Calculates what the dialog width should be depending on title length of the task but fixes it to 700px is it's more than that so it won't grow too big.

if pop-up width is bigger than 450 + title length and 450  title length is less than 700, the dialog width is 450 + title length otherwise it's 400 which is default.

Gnyblast: well specificall this: `this.dialogWidth = (window.innerWidth > 450 + this.templateAlias.length` if windows size can handle the pop-up to growth. That what this line checks. `window.innerWidth` is the size of the screen and it checks if growing pop-up bit more is available on the screen.

Gnyblast: 1. Check is windows size can support growing new task dialog a bit more
2. If yes, then check if title length + 450px is less than 700px (700 can change here, it's a value I think the visual limit of a pop-up page)
3. If yes above two, then make the pop-up grow up to 450px + title length
4. else, make is 400 which is the default defined for `dialogWidth` at `data()` section.

fiftin: Bot `this.templateAlias.length` not in pixels. The result of calculation is incorrect.
I will merge the PR if you remove this fragment.

fiftin (approved):

Gnyblast:
Well you are right, each character length doesn't mean +1 pixel. There are some calculations assuming that in general 1 character is equivalent of 8 pixels. Do you think we can  do `this.templateAlias.length *8`? Because Obviously the pop-up with overflowing to hidden title is disturbing. but I can take it out completely if you are not OK with it.

Gnyblast:
@fiftin I reverted that part about pop-up dialog.

fiftin:
@Gnyblast thank you!

Issue #2733, opened 2025-02-03
Feature: Allow configuring repository directory


vdl-dsi-architecture:
### Related to

Ansible (task execution)

### Impact

must have

### Missing Feature

When creating a repository, an optional field should be the root to use for a specific repository. Infra teams use mono repos or specific structures in which the root of the repository is not where scripts run from.

See #2597, #2531, #2713, #1984, #1515, #1609, #2030

### Implementation

New field on repository, run scripts from that directory.

v-kumar:
Any idea when this will be available?

pb4162:
Commenting as there hasn't been any activity in this issue for over a year. Are there still plans to merge #2766?

fiftin:
Hi @pb4162 it is not so easy as we expected.

We have functionality which check changes in the repo. It is not so easy to implement this for specific folder instance the repo.

PR #2766, opened 2025-02-23
Feature: Ansible directory path inside repo

tjhowse:
Implements #2733. Allows for a repo to be configured with an optional relative path from the Git root to the location inside the repo that stores the playbooks, role directory and configuration.

I'm aware of some blocking tasks:

1. Complete the "isValidPath()" validator,
2. Work out the best place to suffix the path onto the repo location,
3. Update the docs,
~~4. Handle DB migration.~~


tjhowse:
I think the CI needs a cache clear. An earlier mistake in my DB migration script tainted the cache with the new column with incorrect constraints. The CI is trying to migrate an already-migrated DB.

Edit: Maybe not? I think I've just misunderstood how the migration system works. It's OK now.

db/Repository.go
```diff
@@ -31,6 +31,7 @@ type Repository struct {
 	SSHKeyID  int    `db:"ssh_key_id" json:"ssh_key_id" binding:"required" backup:"-"`
 
 	SSHKey AccessKey `db:"-" json:"-" backup:"-"`
+	Path   string    `db:"path" json:"path" backup:"-"`
```
fiftin: `backup:"-"` - must be `backup:"path"`.

fiftin:
Currently `Playbook Filename` can be path to the file. What reason to have `path` field?

Perhaps it is documentation issue?

Perhaps `path` must be in `Repository` table?

samsen1:
what;s blocking this? we have ansible playbooks in repos that holds other code base not related to ansible. this feature is very much needed since we store custom ansible.cfg inside the subfolder and not at the root of the git repo

fiftin:
> what;s blocking this?

Unfixed issue

db/Repository.go
```diff
@@ -44,6 +45,10 @@ func (r Repository) GetDirName(templateID int) string {
 	return r.getDirNamePrefix() + "template_" + strconv.Itoa(templateID)
 }
 
+func (r Repository) GetAnsiblePath(templateID int) string {
```
fiftin: I don't think `GetAnsiblePath` is a correct name, because it can be Terraform or any other repository.

fiftin:
Will be merged to 2.18.

Issue #2862, opened 2025-03-29
Question: How to define variables in the task template?


joaomarcos85:
### Question

I am currently trying to migrate from Ansible AWX to Semaphore. In Ansible AWX I have the option to define variables only for a template. How can I do this in Semaphore?

Below is a screenshot of how I do it in Ansible AWX:
![Image](https://github.com/user-attachments/assets/ec156252-8f78-44a1-81f8-44f69ad4c2d0)

I found the option to create a Variable Groups in Semaphore, but I need to define the variables per template. If I am going to use Variable Groups I will need to create a Variable Group for each template.

Is there a simpler way to do this?

### Related to

Ansible (task execution)

fiftin:
Looks like a useful ability. But currently it is not possible. I think we will add this in 2.14 as part of this feature: https://github.com/semaphoreui/semaphore/issues/2612

joaomarcos85:
> Looks like a useful ability. But currently it is not possible. I think we will add this in 2.14 as part of this feature: [#2612](https://github.com/semaphoreui/semaphore/issues/2612)

@fiftin  I opened a new pull request to add this feature: https://github.com/semaphoreui/semaphore/pull/2872

mickaeltardy:
Hello 
@joaomarcos85 Did you manage to handle the extra variables? 

@fiftin I saw that the pull request was rejected, what is current best practice to specify values of variables in task template? 
I would like to provide values to vars_prompt of my ansible playbook.

fiftin:
@mickaeltardy you should use variable groups

joaomarcos85:
@mickaeltardy @fiftin Currently, Currently, for each template I have to create a variable group. Therefore, when I need to edit the extra variable values of a template, I have to search for the extra variable group and edit it. Since in my case I don’t reuse the same variable group in more than one template, I always need to have the same number of variable groups as the number of templates.

joaomarcos85:
@mickaeltardy @fiftin If at least in the template edit modal there were an option to open the modal for editing the values of the selected variable group, it would already make it much easier to edit the template’s variable values

PR #2872, opened 2025-04-01
feat(template): implements extra variables in the template

joaomarcos85:
Adds the ability to specify extra variables directly in the template, as is done in Ansible AWX:
image

This pull request resolves issue https://github.com/semaphoreui/semaphore/issues/2862.

### DIFF
--- a/api-docs.yml
+++ b/api-docs.yml
@@ -796,6 +796,9 @@ definitions:
         type: integer
       autorun:
         type: boolean
+      extra_vars_json:
+        type: string
+        example: '{}'
 
   Template:
     type: object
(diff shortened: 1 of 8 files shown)


joaomarcos85:
@fiftin Hello, is there still something missing to be done to be merged?

fiftin:
Hi @joaomarcos85,

This feature will be implemented other way. Sorry, I can't merge this PR. This solution is not universal.

It will be implemented using Variable Group associated with Template.

As I already mentioned, it will be done, when multiple variable groups for Template will be implemented.

joaomarcos85:
@fiftin  Thanks a lot for the feedback! What is the purpose of using multiple variable groups for the template? 
What I am looking for (and implementing in this PR) is the option to define variables only for a single template, without having to create a variable group for each template. Some variables are only used in one template, but in another template they would have another value. Would the functionality of multiple variable groups for the template solve this problem?

PR #2918, opened 2025-04-14
Feature: populate TaskParams from integration extract value

MIJohnson:
Allows Extract Value from integrations to be injected into the executing Task allowing users to control external properties like limit

When creating an Extract Value, when selecting 'Environment' everything will work as usual. If selecting Task, the value(s) will be injected into the Task.Params

extractval3
extractval2
extractval1


fiftin:
Hi @MIJohnson the PR looks very solidity, thank you :)

It looks very well for me.

I trying to understand what happened with Codacy.

MIJohnson:
Perhaps the codacity issues related to these? image

fiftin:
I didn't find a reason to not merge it, thank you :)

billglick:
Can you walk me through how to use this? 

I'm currently using Semaphore UI 2.16.51 which appears to have this "Task Params" patch.

I'm having difficulty making use of this with ExctractValue type task. I have a JSON BODY Value named `host` that contains a string like "exampleShortHostName*". Then I'm setting my ExtractValue as:
- Name: limit
- Value Source: body
- Data Type of Body: JSON
- Key: host
- Variable Usage: Task Params
- Variable: limit

My old method of this was to use an ExtractValue with an environment Variable named "TARGET_HOST"  and then set my Ansible playbook up to use the following:
```yaml
---
- name: "Base configuration"
  # FOLLOWING IS A HACK TO LET US TARGET A SPECIFIC HOST VIA WEBHOOK INTEGRATION
  hosts: "{{ TARGET_HOST | default('all') }}"
  ...
```
This approach works, but is a bit annoying that I have to edit my playbook. But I mention this to say that I'm definitely getting the "host" JSON into the BODY of my web hook correctly. I just cannot figure out how to extract it directly into my `--limit` like you describe.

PR #2920, opened 2025-04-15
fix: use md5 files for each GalaxyRequirementsType

btravouillon:
This solves an issue in installGalaxyRequirementsFile() where only the collections are installed while the requirements.yml file includes both collections and roles. This function is called to install both Galaxy requirements types, however it checks if the requirements has changed before installing.

One the collections were installed, a new call will skip roles installation because the requirements file has not changed since.

Bug observed with v2.13.12


fiftin:
Thanks a lot!

Issue #3027, opened 2025-05-14
Problem: Incorrect redirection with `web_host` in server configuration under Templates tab.


MichalTomczakSE:
When attempting to add a new application via the "Templates" section while using the `web_host` parameter in server configuration (e.g., `/semaphore`), the redirection URL is incorrect. Instead of navigating to the expected path `/<web_host>/apps` (e.g., `/semaphore/apps`), the user is redirected directly to `/apps`.

**Steps to Reproduce:**

1. Navigate to the "Templates" section for adding a new application.
2. Ensure that the application is being accessed with a `web_host` parameter defined (e.g., accessing the application via a URL like `http://example.com/semaphore`).
3. Click on the button or link to add a new application.

**Expected Behavior:**

The user should be redirected to a URL that includes the `web_host` prefix, specifically `/<web_host>/apps`. For example, if `web_host` is set to `/semaphore`, the expected redirection URL should be `/semaphore/apps`.

**Actual Behavior:**

The user is redirected to `/apps`, which is missing the `web_host` prefix. This likely results in an error or an incorrect view within the application.

### Impact

Web-Frontend (what users interact with)

### Installation method

Binary

### Database

BoltDB

### Browser

Microsoft Edge

### Semaphore Version

2.14.8-121aa85-1746383708
develop-d8429412-1747204333

### Ansible Version

```bash

```

### Manual installation - system information

RHEL 4.18.0-553.47.1.el8_10.x86_64 #1 SMP Thu Mar 20 10:01:53 EDT 2025 x86_64 x86_64 x86_64 GNU/Linux
Rocky Linux 5.14.0-503.38.1.el9_5.x86_64 #1 SMP PREEMPT_DYNAMIC Wed Apr 16 16:38:39 UTC 2025 x86_64 x86_64 x86_64 GNU/Linux

### Configuration

{
        "bolt": {
                "host": "/semaphore/database.boltdb"
        },
        "dialect": "bolt",
        "tmp_path": "/tmp/semaphore",
        "web_host": "/semaphore",
        "cookie_hash": "x",
        "cookie_encryption": "x",
        "access_key_encryption": "x"
 }

PR #3028, opened 2025-05-14
fix: : Correct redirection with `web_host` to /apps endpoint #3027

MichalTomczakSE:
## Fix: Correct Redirection with `web_host` #3027

Previously, when a `web_host` (e.g., `/semaphore`) was used, the redirection would incorrectly navigate directly to `/apps` instead of the expected `/<web_host>/apps` (e.g., `/semaphore/apps`).

**Reasoning for the Change:**

The previous implementation used a standard `href` attribute for the link, which always resulted in a full page load and navigation to the absolute path `/apps`, ignoring the `web_host` context.

This change replaces the `href` attribute with the Vue Router's `to` prop. By utilizing `to`, the Vue Router handles the navigation internally, correctly resolving the target path relative to the application's base URL, including the `web_host` if it is defined.

**Testing:**

- Verified that when accessing the application without a `web_host`, the redirection to `/apps` remains correct.
- Confirmed that when accessing the application with a `web_host` (e.g., `/semaphore`), the redirection now correctly navigates to `/<web_host>/apps` (e.g., `/semaphore/apps`).

I was not able to run the dredd tests on my environment (I am really sorry for that). I am also not sure if this change could have impact to other things (I believe it should not).

### DIFF
--- a/web/src/views/project/Templates.vue
+++ b/web/src/views/project/Templates.vue
@@ -90,7 +90,7 @@
             v-if="isAdmin"
             key="other"
             link
-            href="/apps"
+            to="/apps"
           >
             <v-list-item-icon>
               <v-icon>mdi-cogs</v-icon>


fiftin:
Thank you!

Issue #2322, opened 2024-08-31
Passing secret variable to Terraform/openTofu task not working


fm-knopfler:
Another user also suspects this is a bug. 

### Discussed in https://github.com/semaphoreui/semaphore/discussions/2287

<div type='discussions-op-text'>

<sup>Originally posted by **fm-knopfler** August 16, 2024</sup>
I set the following survey variable as a secret.
![grafik](https://github.com/user-attachments/assets/97087742-a76b-41f3-be04-024df4acb0f2)

And declared it like this in my terraform code.
```terraform
variable "api_token" {
  type      = string
  sensitive = true
}
```

Still, I get this error. With pure string variables (like hostname) it works fine. But as soon as I define it as a secret in either Semaphore or my TF code (or both), I get this error.
```
3:31:08 PM If you ever set or change modules or backend configuration for OpenTofu,
3:31:08 PM rerun this command to reinitialize your working directory. If you forget, other
3:31:08 PM commands will detect it and remind you to do so if necessary.
3:31:08 PM var.api_token
3:31:08 PM╷
3:31:08 PM│ Error: No value for required variable
3:31:08 PM│ 
3:31:08 PM│   on variables.tf line 11:
3:31:08 PM│   11: variable "api_token" {
3:31:08 PM│ 
3:31:08 PM│ The root module input variable "api_token" is not set, and has no default
3:31:08 PM│ value. Use a -var or -var-file command line argument to provide a value for
3:31:08 PM│ this variable.
```

Am I doing something wrong or is this a bug? </div>

bencurio:
Hi,

If you want to pass it as an environment variable, you need adding the `TF_VAR_` prefix, without it it will not work.

So for example in your case:

`TF_VAR_api_token=your_token`

Please correct me, but I think secrets is added as an environment variable.

Related docs: https://developer.hashicorp.com/terraform/language/values/variables#environment-variables

mbusiek-nasuni:
I'm running into this as well. 

I have tried setting the variable name as `TF_VAR_variable-name`, but it still doesn't seem to get passed to terraform/opentofu. 

for example:

![Image](https://github.com/user-attachments/assets/a0e5f992-ce36-4423-9ab6-b560423bbac5)

this configuration for my secret key variable does not work to pass the variable to the script at all. I'm not sure if it's a type error or something, but no variable is created by that name in terraform/opentofu

if I set it to a string type, it passes just fine to the script.

[...]

PR #3083, opened 2025-06-07
fix: add secret survey variable on opentofu task issue #2322

bdoubrovine:
This pull request introduces enhancements to the `getEnvironmentExtraVars` method in `services/tasks/LocalJob.go` to improve the handling of secret variables. The most important change involves merging secret variables into the existing environment variables while ensuring the secret is reset after processing.

### Enhancements to secret variable handling:

* Added a new map, `extraSecretVars`, to handle secret variables separately and parse them from the `t.Secret` JSON string. If parsing fails, the method returns an error.
* Reset the `t.Secret` field to an empty JSON object (`"{}"`) after processing to ensure that secrets are not retained in memory.
* Merged the parsed `extraSecretVars` into the `extraVars` map using `maps.Copy`, consolidating all variables into a single map for further use.

#2322 and #2287

### DIFF
--- a/services/tasks/LocalJob.go
+++ b/services/tasks/LocalJob.go
@@ -67,6 +67,7 @@ func (t *LocalJob) SetCommit(hash, message string) {
 func (t *LocalJob) getEnvironmentExtraVars(username string, incomingVersion *string) (extraVars map[string]any, err error) {
 
 	extraVars = make(map[string]any)
+	extraSecretVars := make(map[string]any)
 
 	if t.Environment.JSON != "" {
 		err = json.Unmarshal([]byte(t.Environment.JSON), &extraVars)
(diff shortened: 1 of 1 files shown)


fiftin:
@cursor review

opipenbe:
Hi! What is the status of this fix? Would be really useful if secrets can be passed as variables.

bdoubrovine:
> Hi! What is the status of this fix? Would be really useful if secrets can be passed as variables.

Hi, this PR is only an add missing code for use survey secret in OpenTofu Task.(copy/paste from ansible secret survey)
On my side, I used it to get credentials need to play a OpenTofu task.

Without this PR, secret suvey variable don't work on Semaphore UI.

PR #2324, opened 2024-09-01
Allow to configure OIDC providers from environment variables

subuk:
Hello!

Small change to allow to specify OIDC providers configuration from environment variables

Docs pr: https://github.com/semaphoreui/semaphore-docs/pull/51/

PR #3117, opened 2025-07-01
Add support for reading OIDC client_secret from file path

ghost:
Currently you can set the SEMAPHORE_OIDC_PROVIDERS environment variable as outlined here:
https://github.com/semaphoreui/semaphore/pull/2324

This PR now supports direct file paths for values, so secrets do not need to be hardcoded:

```
SEMAPHORE_OIDC_PROVIDERS: '{
"authentik": {
"display_name": "Sign in with Authentik",
"provider_url": "https://authentik.example.com/application/o//",
"client_id": "semaphore",
"client_secret": "/path/to/secret",
"redirect_url": "https://semaphore.example.com/api/auth/oidc/authentik/redirect/",
"scopes": ["openid", "profile", "email"],
"username_claim": "preferred_username",
"name_claim": "preferred_username"
}'
}
```

Rather simply, this pulls the config with the value as a string, then checks if it is a file path and overwrites it with the file contents if valid and non-empty.

### DIFF
--- a/util/config.go
+++ b/util/config.go
@@ -384,6 +384,7 @@ func ConfigInit(configPath string, noConfigFile bool) (usedConfigPath *string) {
 
 	loadConfigEnvironment()
 	loadConfigDefaults()
+	loadClientSecretsFromFiles(Config.OidcProviders)
 
 	fmt.Println("Validating config")
 	validateConfig()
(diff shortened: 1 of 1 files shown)


ghost:
This functionality was already built in via another file. Apologies!

PR #3159, opened 2025-07-22
Support task params for Schedules/Integations

fiftin:
- **feat(ui): refactor task form**
- **refactor(tasks): move ansible params to separate component**
- **feat(tasks): add task params table**
- **feat(schedule): support task params**
- **feat(integrations): save task settings**
- **feat(schedule): spaces**
- **fix(schedule): cron builder**
- **fix(schedule): sql query**
- **fix(tasks): override env via survay variables**
- **feat(schedule): use task params**
- **fix(integrations): empty task params**


naimo84:
Great idea and very useful. Might it be possible to set the hosts limit in the schedule? Or is it already implemented and do i miss some setting? :)

image

PR #3207, opened 2025-08-16
refactor: replace []byte(fmt.Sprintf) with fmt.Appendf

yinwenyu6:
This change enhances performance by directly appending the formatted string to the byte slice, avoiding the need for intermediate memory allocation created by fmt.Sprintf. 
 
 The resulting code is cleaner and more efficient, maintaining the same functionality while improving resource usage.
 
 More info can see https://github.com/golang/go/issues/47579


yinwenyu6:
@fiftin Hi, Could you please review this PR at your convenience? Thank you very much.

PR #3244, opened 2025-08-28
feat(ui) submit edit dialogs on enter

apollo13:
I don't really know vue so I am not sure this is correct :)


fiftin:
@apollo13 Some dialogs have multiline text inputs, so I don't think this behavior works well for all dialogs.

apollo13:
Good point, will check. But even then we should find a way to not reload the page on enter like it currently does (not exactly sure though why, I am not really a frontend JavaScript person)

PR #3301, opened 2025-09-15
fix: improve the openapi.yml file

bluntelk:
Testing against https://github.com/daveshanley/vacuum I have improved the quality from 10% to 25% by
 - adding operationId fields to each endpoint
 - fixed an error that prevented the open api generator from generating code
 - fixed the tags to be consistent


fiftin:
Thank you!

PR #3384, opened 2025-10-29
Force galaxy requirements install via environment variable.

supersaiyanmode:
This is especially useful when 1) the contents of the requirements file doesn't really change, 2) versions are not being used.


supersaiyanmode:
PLMK if this env var  set via a task template would take effect here. Thanks! :)

Issue #3171, opened 2025-07-29
Feature: Add SSH Cert to Keystore


ephemeric:
### Related to

Web-Frontend (what users interact with)

### Impact

must have

### Missing Feature

The per user ability to add a SSH certificate to the `Key Store` alongside an existing `ssh` key type.

### Implementation

The UI should have a text box field to add a SSH cert when creating a `ssh` type key in the store.

<img width="1440" height="900" alt="Image" src="https://github.com/user-attachments/assets/123b4554-7a0c-4c7e-8fbd-f5cea1bb77eb" />

### Design

The UI should provide an element to facilitate the add SSH certificate requirement.

I have a workaround in place that uses `--ssh-extra-args="-o CertificateFile=/home/robertg/semcert"` and requires the user homedir to be `0750` and group `semaphore`. See screenshot.

The systemd service unit file has User and Group as semaphore:semaphore respectively.

The `semaphore` user homedir is in `/var/lib/semaphore` and adding a cert in `/var/lib/semaphore/.ssh/id_ed25519-cert.pub` does work but does not fulfill per user requirements and would fail an audit. Was mentioned [here](https://github.com/semaphoreui/semaphore/issues/271) but remains illucid as to exactly what is/is not supported and I couldn't find anything in the docs either.

<img width="1440" height="900" alt="Image" src="https://github.com/user-attachments/assets/03748d85-2ea8-4e90-9c41-060d6fb56e0c" />

PR #3531, opened 2025-12-16
SSH CA certificate key store support

LucasFrendorf:
Allows users to add SSH CA certificate to SSH key in key store

image

This solves https://github.com/semaphoreui/semaphore/issues/3171


fiftin:
Hi @LucasFrendorf the PR already exists. We have `plain` field for this, we don't need new one.

fiftin:
https://github.com/semaphoreui/semaphore/pull/3656

LucasFrendorf:
Hi @fiftin, this PR is about SSH CA certificate support (#3171), a CA cert is a signed public key that gets passed alongside the SSH private key during connection via -o CertificateFile=, so the remote host can verify the key was signed by a trusted CA (This isn't covered by #3656)
I can move the cert storage to the plain field, though #3656 would currently wipe it
But i would need to tweak that to allow the cert through, does that sound good?
If so i will wait for #3656 to merge

PR #3536, opened 2025-12-18
Multi-repository task support

piotrkochan:
This change allows tasks to clone multiple git repositories alongside the main repository during execution. Additional repos are cloned into `<workspace>/repos/` subdirectory with configurable paths.

 Use case: Projects requiring dependencies from multiple repositories (e.g., shared libraries, configuration repos, get project code to build or external modules) can now be automatically cloned and available during task execution without manual scripting.

image

image


piotrkochan:
Much better solution https://github.com/semaphoreui/semaphore/pull/3547

PR #3554, opened 2026-01-04
feat: Add Lark webhook notification medium

shaxiaozz:
Hello, we need to use Lark webhook notification, so we added the code of this notification medium, hoping to merge it in. Because we will push the platform to the production environment for formal use. Thank you~

This code has been tested and verified in the local development environment, and the following is the effect diagram:


shaxiaozz:
Hi @fiftin 
Can you help review and merge the code?

PR #3695, opened 2026-03-12
Add admin permission check for project restore from backup

JulianKap:
## PR Title

Add admin permission check for project restore from backup

## Description

This PR adds a missing **admin permission check** when restoring a project from backup.

## Problem

In the current implementation, creating a new project correctly requires **admin privileges**.
However, the **project restore from backup** endpoint did not perform the same permission check.

As a result, a non-admin user could restore a project from a backup and effectively create a new project without having the required permissions.

This created a permission bypass scenario:

* non-admin users **cannot create projects directly**
* but they **could restore a project from backup**, which results in a new project being created

## Solution

This PR adds the same **admin permission validation** used in project creation to the **project restore** operation.

Now:

* only users with **admin privileges** can restore a project from backup
* permission checks are consistent with the normal project creation flow

## Result

* Prevents permission bypass via project restore

### DIFF
--- a/api/projects/backup_restore.go
+++ b/api/projects/backup_restore.go
@@ -39,7 +39,7 @@ func Restore(w http.ResponseWriter, r *http.Request) {
 	user := helpers.GetFromContext(r, "user").(*db.User)
 
 	if !user.Admin && !util.Config.NonAdminCanCreateProject {
-		log.Warn(user.Username + " is not permitted to to restore the project")
+		log.Warn(user.Username + " is not permitted to restore the project")
 		w.WriteHeader(http.StatusUnauthorized)
 		return
 	}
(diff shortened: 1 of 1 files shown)


api/projects/backup_restore.go
```diff
@@ -37,6 +38,12 @@ func GetBackup(w http.ResponseWriter, r *http.Request) {
 func Restore(w http.ResponseWriter, r *http.Request) {
 	user := helpers.GetFromContext(r, "user").(*db.User)
 
+	if !user.Admin && !util.Config.NonAdminCanCreateProject {
+		log.Warn(user.Username + " is not permitted to edit users")
```
fiftin: It looks like this message incorrect

JulianKap: Hi @fiftin , I updated the message text in the log.

PR #3705, opened 2026-03-17
fix(i18n): improve zh_TW translation consistency

tan9:
## Summary

Follow-up to #3701. This PR addresses remaining consistency issues in Traditional Chinese (zh_TW) translations:

- **Fix environment-related translations**: The key `environment` maps to "Variable Groups" in English, but some related keys (`askDeleteEnv`, `enterEnvJson`, `environment_required`) were incorrectly using "環境" (environment) instead of "變數群組" (variable group). Now uses "變數群組" for Variable Group context, and "環境變數" for actual environment variables.
- **Standardize "新xx" to "新增xx"**: All action-oriented labels like "New Project", "New Template", "New Key" etc. now consistently use "新增" (add/create) prefix instead of bare "新" (new), which is the standard convention in Taiwan UI design.
- **Keep "Playbook" untranslated**: Replaced "劇本" with "Playbook" as it is a well-known Ansible technical term that Taiwan users expect to see in English.
- **Standardize "覆蓋" to "覆寫"**: Use "覆寫" consistently for "override" across all translations.
- **Standardize "（可選）" to "（選填）"**: Use "選填" to pair naturally with "必填" (required), following Taiwan form UI conventions.
- **Fix "表達式" to "表示式"**: Use the standard Taiwan term for "expression".

## Test plan

- [x] Switch UI language to Traditional Chinese and verify translated strings on all pages
- [x] Confirm "新增" prefix appears consistently on all "New ..." dialog titles and buttons
- [x] Verify environment / variable group related pages use correct terminology


fiftin:
Hi, thank you!

PR #3725, opened 2026-03-28
feat(ui): Add quick search filter to the Task Templates view

akeelow:
## Description
When working with a large number of Task Templates, manually scrolling through them becomes tedious and inefficient.
This PR adds a quick, client-side search input filter to the Task Templates table. 

## Changes Made
- Added a `<v-text-field>` in the `Templates.vue` toolbar for the search input.
- Bound the search query to the existing `v-data-table`'s built-in search property.
- Added translation key `search` to `en.js` and `ru.js` localization files.

## How to Test
1. Navigate to navigating to the `Task Templates` section.
2. Type in the text field at the top right of the toolbar.
3. The list is automatically filtered locally based on query criteria.

### DIFF
--- a/web/src/lang/de.js
+++ b/web/src/lang/de.js
@@ -214,6 +214,7 @@ export default {
   editViews: 'Ansichten bearbeiten',
   newTemplate: 'Neue Vorlage',
   taskTemplates2: 'Task-Vorlagen',
+  search: 'Suche',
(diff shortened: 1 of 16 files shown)


heshc:
I support this, looking forward to getting it merged into the latest release quickly.

PR #3757, opened 2026-04-09
feat: Add a 'pull submodules' option to Repositories

stith:
Our Semaphore instance uses a monorepo as its Ansible playbooks source, and that monorepo has several large submodules that we don't care about when doing Ansible runs, which adds ~5 minutes to each new clone. This PR adds a "Pull submodules" option (defaulting to `true` to match current behavior) which lets us turn that behavior off.

This is my first time contributing to this project and I'm not sure how to decide which version number to use for the migrations, so advice is welcome!


utafrali (commented):
The feature is well-scoped and covers both git client implementations (CmdGitClient and GoGitClient) consistently. The main issues are: the migration column type (`int` vs `boolean`) and MySQL-specific backtick syntax that may break PostgreSQL support, duplicated code in `CmdGitClient.Clone`, and a subtle Go zero-value default that could silently disable submodules when the struct is constructed outside the frontend path.

db/sql/migrations/v2.18.0.sql
```diff
@@ -0,0 +1 @@
+alter table `project__repository` add `pull_submodules` boolean not null DEFAULT true;
```
utafrali, line 1: Two issues here:

1. The column type is `int` but the Go struct field is `bool`. Other parts of the codebase use `boolean` (or `tinyint(1)`) for boolean columns, which maps more naturally to Go's `bool`. Change to:
```sql
alter table `project__repository` add `pull_submodules` boolean not null default true;
```

2. The backtick syntax is MySQL-specific. Since `db/Migration.go` accepts a `dialect` parameter, the project likely supports multiple databases (e.g. PostgreSQL). Check whether this migration needs a PostgreSQL variant, or whether the project has a mechanism to handle dialect differences for migrations.

db/sql/migrations/v2.18.0.err.sql
```diff
@@ -0,0 +1 @@
+alter table `project__repository` drop column `pull_submodules`;
```
utafrali, line 1: Same concern as the forward migration: backtick syntax is MySQL-specific and will fail on PostgreSQL. If the project supports multiple dialects, this rollback file needs to handle that too.

db_lib/CmdGitClient.go
```diff
@@ -100,9 +100,17 @@ func (c CmdGitClient) Clone(r GitRepository) error {
 		dirName = r.TmpDirName
 	}
 
+	if r.Repository.PullSubmodules {
```
utafrali, line 103: The `Clone` function now has two nearly identical `c.run(...)` calls that diverge only by the presence of `--recursive`. This is harder to maintain and easy to accidentally desync. The `GoGitClient` approach is better -- build the args list conditionally before calling run:

```go
args := []string{"clone", "--branch", r.Repository.GitBranch, r.Repository.GetGitURL(false), dirName}
if r.Repository.PullSubmodules {
    args = append([]string{"clone", "--recursive", "--branch", r.Repository.GitBranch, r.Repository.GetGitURL(false), dirName}, []string{}...)
    // or build args slice and insert --recursive at position 1
}
return c.run(r, GitRepositoryTmpPath, args...)
```

A cleaner approach:
```go
cloneArgs := []string{"clone"}
if r.Repository.PullSubmodules {
    cloneArgs = append(cloneArgs, "--recursive")
}
cloneArgs = append(cloneArgs, "--branch", r.Repository.GitBranch, r.Repository.GetGitURL(false), dirName)
return c.run(r, GitRepositoryTmpPath, cloneArgs...)
```

db_lib/GoGitClient.go
```diff
@@ -137,9 +142,16 @@ func (c GoGitClient) Pull(r GitRepository) error {
 	}
 
 	// Pull the latest changes from the origin remote and merge into the current branch
-	err = wt.Pull(&git.PullOptions{RemoteName: "origin",
+	pullOpts := &git.PullOptions{
```
utafrali, line 145: Minor redundancy: `RecurseSubmodules` is set to `git.DefaultSubmoduleRecursionDepth` in the struct literal, then immediately overridden to `git.NoRecurseSubmodules` if `!PullSubmodules`. It reads more clearly to initialize it to `git.NoRecurseSubmodules` and only set the recursive depth when needed -- matching the pattern used in the `Clone` method above:

```go
pullOpts := &git.PullOptions{
    RemoteName:        "origin",
    Auth:              authMethod,
    RecurseSubmodules: git.NoRecurseSubmodules,
}
if r.Repository.PullSubmodules {
    pullOpts.RecurseSubmodules = git.DefaultSubmoduleRecursionDepth
}
```

db/Repository.go
```diff
@@ -22,12 +22,13 @@ const (
...
-	ProjectID int    `db:"project_id" json:"project_id" backup:"-"`
-	GitURL    string `db:"git_url" json:"git_url" binding:"required"`
-	GitBranch string `db:"git_branch" json:"git_branch" binding:"required"`
-	SSHKeyID  int    `db:"ssh_key_id" json:"ssh_key_id" binding:"required" backup:"-"`
+	ID             int    `db:"id" json:"id" backup:"-"`
+	Name           string `db:"name" json:"name" binding:"required"`
+	ProjectID      int    `db:"project_id" json:"project_id" backup:"-"`
+	GitURL         string `db:"git_url" json:"git_url" binding:"required"`
+	GitBranch      string `db:"git_branch" json:"git_branch" binding:"required"`
+	SSHKeyID       int    `db:"ssh_key_id" json:"ssh_key_id" binding:"required" backup:"-"`
```
utafrali, line 30: The zero value of `bool` in Go is `false`, so any code path that constructs a `Repository` struct without explicitly setting `PullSubmodules` (e.g. direct API calls, backup restore, or internal tooling) will silently disable submodule pulling. The database default of `1` only applies to rows inserted without that column, not to the application layer.

Consider whether the API handler that reads the request body into a `Repository` struct needs to default this field to `true` when it is absent from the JSON payload. A JSON `omitempty` tag would also prevent the frontend from explicitly sending `false` when unchecked. At minimum, this edge case is worth a comment.

web/src/components/RepositoryForm.vue
```diff
@@ -174,6 +180,10 @@ export default {
       this.helpDialog = true;
     },
 
+    getNewItem() {
```
utafrali, line 183: If the base component (`ItemFormDialog` or whatever this extends) already defines a `getNewItem()` method, this override silently drops any defaults the parent sets. Check whether you need to merge with the parent:

```js
getNewItem() {
  return { ...super.getNewItem?.() ?? {}, pull_submodules: true };
},
```

This also ensures forward compatibility if the parent gains new defaults later.

db/sql/migrations/v2.18.0.sql
stith: who are you

stith:
is this the official review bot for this project? it doesn't seem to have any past activity. what's up with that?

fiftin:
Hi @stith no, it is not official review bot.

stith:
what's the process for review on this? do i need to open an issue for discussion first?

fiftin:
I'm not sure this flag should be in repository table. The template table looks more suitable for this.

PR #3788, opened 2026-04-17
feat: add name field to API tokens

setswei:
- Add Name field to APIToken model
- Add v2.17.18 SQL migration to add name column to user__token table
- Update SQLite bootstrap migration to include name column
- Update createAPIToken API handler to accept optional name in request body
- Add name column to tokens table in frontend
- Add dialog prompt for token name when creating new tokens


setswei:
I have been using semaphoreui for quite some time now, and i am starting to use a lot of API tokens to allow other tools to integrate with it. I follow strict security practices within the environments i work with, and one of these is that each integration needs to have its own credentials, etc. 

So i added a name field to the API Tokens section that allows you to add a descriptive name to the token. all tests, db migration files have been updated to support this.

image

fiftin:
Thank you!

PR #3831, opened 2026-05-05
fix(api): prevent 404 after OIDC login when web_host is empty

### LINKED ISSUE #2681: Problem: 404 on successful login via OIDC if web_host config option is empty
### Issue

**Problem**
If Semaphore is hosted without a subpath such as `https://semaphoreserver.local:9999/` then according to the [config docs](https://docs.semaphoreui.com/administration-guide/configuration/) the relevant option for the setting `web_host` should not have a trailing slash and should be empty in the config.

When logging in via OIDC (and perhaps other OAuth mechanisms) a successful auth flow returns the user to `https://https://semaphoreserver.local:9999/api/auth/oidc/oidcprovidername/` with a HTTP 404 error and fails to redirect them to the Semaphore UI. Navigating away from this page to a URL such as `https://semaphoreserver.local:9999/project/1/history` works, which shows that the user has logged in successfully.

**Cause & Workaround**
After changing the `web_host` config value to "/" and restarting Semaphore, the user is now properly redirected upon a successful auth flow. I am not sure if this has any other negative impacts.

_This behaviour could potentially be the cause for issues #1434, #2659, and #2666👿._

### Impact

Web-Frontend (what users interact with), Web-Backend (APIs)

### Installation method

Package

### Database

BoltDB

### Browser

Chrome

### Semaphore Version

2.11.2-0e9490c-1735214524

### Ansible Version

```bash
Not relevant.
```

### Logs & errors

_No response_

### Manual installation - system information

OS: Debian
Reverse proxy: Nginx

### Configuration

_No response_

### Additional information

_No response_

SAY-5:
Fixes #2681.

When `web_host` is unset (Semaphore mounted at root, per the [config docs](https://docs.semaphoreui.com/administration-guide/configuration/)), `url.JoinPath("", "/project/1")` returns `"project/1"`, the leading slash is stripped, producing a relative URL. The browser resolves it against the current `/api/auth/oidc/<id>/redirect` URL, landing the user at `/api/auth/oidc/<id>/project/1` and returning 404. The reporter's workaround of setting `web_host = "/"` confirms the cause.

Extracted the redirect-URL build into `buildOidcRedirectURL` and return an absolute path beginning with `/` when `web_host` is empty. Added table-driven tests for empty and configured `web_host` cases.

Tested with `go test ./api/... -run Login`.

### DIFF
--- a/api/login.go
+++ b/api/login.go
@@ -809,20 +809,40 @@ func oidcRedirect(w http.ResponseWriter, r *http.Request) {
 		redirectPath = mux.Vars(r)["redirect_path"]
 	}
 
-	if !strings.HasPrefix(redirectPath, "/") {
-		redirectPath = "/" + redirectPath
-	}
-
-	redirectURL, err := url.JoinPath(util.Config.WebHost, redirectPath)
+	redirectURL, err := buildOidcRedirectURL(util.Config.WebHost, redirectPath)
 	if err != nil {
 		log.Error(err)
 		http.Redirect(w, r, loginURL, http.StatusTemporaryRedirect)
 		return
 	}
 
+	http.Redirect(w, r, redirectURL, http.StatusTemporaryRedirect)
+}
+
+// buildOidcRedirectURL builds the post-OIDC-login redirect target.
+//
+// When webHost is empty (Semaphore mounted at root, per docs), url.JoinPath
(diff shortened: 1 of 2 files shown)


SAY-5:
Closing, this landed on develop as oidcSuccessRedirectURL in eb42b2a with the same empty web_host handling, so there's nothing left here.

PR #3906, opened 2026-05-31
Adds support for JSON survey variables so task templates can define survey inputs whose values are stored and submitted as structured JSON instead of plain strings.

jbarcia2019:
## Changes
  - Adds a new `json` survey variable type.
  - Allows `survey_vars[].default_value` to contain any valid JSON value.
  - Adds JSON editors and validation when editing survey variables and launching tasks.
  - Updates the API documentation for the new survey variable type.
  - Keeps existing string, integer, enum, and secret survey variables compatible.

  ## Compatibility
  No database migration is required because survey variables are already stored as a JSON blob in the existing `project__template.survey_vars` text column. Existing string defaults remain valid and deserialize
  unchanged.

  ## Testing
  - Ran `git diff --check`.
  - Built successfully through the local Docker image build.

### DIFF
--- a/api-docs.yml
+++ b/api-docs.yml
@@ -1047,10 +1047,12 @@ definitions:
         type: string
       type:
         type: string
-        enum: ["", int, enum, secret] # String => "", Integer => "int"
+        enum: ["", int, enum, secret, json] # String => "", Integer => "int", JSON => "json"
         example: int
       required:
         type: boolean
+      default_value:
+        description: Default survey value. JSON survey variables may use any valid JSON value.
       values:
         type: array
         items:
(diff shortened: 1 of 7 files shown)


jbarcia2019:
The failing check appears unrelated to this PR.

  The `Add PRO implementation` step in `.github/workflows/dev.yml` clones `semaphoreui/semaphorepro-module` using `secrets.GH_TOKEN`. Since this PR comes from a fork, GitHub Actions does not expose repository
  secrets to the `pull_request` workflow, so the clone fails with exit code 128.

  The change itself only touches the task templates page search UI and translations.

@fiftin

Issue #3693, opened 2026-03-11
Feature: Pushover Support


declan727:
### Related to

Other

### Impact

nice to have

### Missing Feature

adding pushover as a messenger would be a huge help and it is also free to use.

https://pushover.net/

### Implementation

It would be a messenger

sickwell6988:
Tried to implement it here: https://github.com/semaphoreui/semaphore/pull/3976

PR #3976, opened 2026-06-17
Add Pushover notifications support (#3693)

sickwell6988:
Originally requested in #3693

Tested and verified

Screenshot 2026-06-18 at 00 06 14

telegram-cloud-photo-size-2-5303018484725915965-y

Tested in Docker

Performed tests agains the new feature, by defining setting in
- config.json
- environment variables
- new Pushover token UI field

Pushover API token UI field > config.json || environment variables


services/tasks/alert.go
```diff
@@ -190,6 +196,77 @@ func (t *TaskRunner) sendTelegramAlert() {
...
+
+	if t.Template.SuppressSuccessAlerts && t.Task.Status == task_logger.TaskSuccessStatus {
+		return
+	}
+
+	token := util.Config.PushoverAPIToken
+	user := util.Config.PushoverUserGroupKey
+
+	if token == "" || user == "" {
+		return
```
wahyudibo, line 212: I think we should put a `t.Log` here so it's easier to debug in case `util.Config.PushoverAlert` is set to true but token or user is empty.

services/tasks/alert.go
```diff
@@ -190,6 +196,77 @@ func (t *TaskRunner) sendTelegramAlert() {
...
+			ID:      strconv.Itoa(t.Task.ID),
+			URL:     t.taskLink(),
+			Result:  t.Task.Status.Format(),
+			Version: version,
+			Desc:    t.Task.Message,
+		},
+		Pushover: alertPushover{Token: token, User: user},
+	}
+
+	tpl, err := template.ParseFS(templates, "templates/pushover.tmpl")
```
wahyudibo, line 232: Since the pushover message is sent as JSON body, i don't recommend using a template for various reasons, from a practical to a security perspective

1. Using struct and JSON marshal is simpler to implement than using a template. We can pass a struct like

```
type pushoverPayload struct {
	Token    string `json:"token"`
	User     string `json:"user"`
	Title    string `json:"title"`
	Message  string `json:"message"`
	URL      string `json:"url,omitempty"`
	URLTitle string `json:"url_title,omitempty"`
}
``` 
JSON-marshall it and pass it to the request body.
2. Performance-wise, parsing a template is more expensive than JSON marshalling. 
3. The JSON body will break if any of the template values contain double quotes, since it's not parsed automatically by the template because it's a plain text template. JSON marshalling fixes this issue
4. As mentioned by cursor [comment](https://github.com/semaphoreui/semaphore/pull/3976/changes#r3432953510) above, using a template for JSON body opens a vulnerability due to JSON injection

--- 

Another thing is the `message` field, which contains HTML text for its value and `html` field is set to `1`.
```
"message": "<b>{{ .Task.Result }}</b> {{ .Task.Version }} - {{ .Task.Desc }}\nby {{ .Author }}",
``` 

Since the value also expects input from the user, it's a security-aware practice to sanitize the input first:
```
message := fmt.Sprintf("<b>%s</b> %s - %s\nby %s",
	html.EscapeString(t.Task.Status.Format()),
	html.EscapeString(version),
	html.EscapeString(t.Task.Message),
	html.EscapeString(author),
)
```

services/tasks/alert.go
```diff
@@ -190,6 +196,77 @@ func (t *TaskRunner) sendTelegramAlert() {
...
+
+	resp, err := http.Post(
+		"https://api.pushover.net/1/messages.json",
+		"application/json",
+		body,
+	)
+
+	if err != nil {
+		t.Log("Can't send pushover alert! Error: " + err.Error())
+	} else if resp.StatusCode != 200 {
```
wahyudibo, line 259: i'm being nitpicky here, but i think we can change hardcoded `200` with `http.StatusOK` which can be understandable at a glance

services/tasks/alert.go
```diff
@@ -190,6 +196,77 @@ func (t *TaskRunner) sendTelegramAlert() {
...
+		t.Log("Can't send pushover alert! Response code: " + strconv.Itoa(resp.StatusCode))
```
wahyudibo, line 260: We can use `t.Logf` to avoid using `strconv.Itoa`
```
t.Logf("Can't send pushover alert! Response code: %d", resp.StatusCode)
```

PR #4007, opened 2026-07-05
Dynamic playbook selection from repository

rici4kubicek:
- Replace the free-text "Playbook" field in Task Templates with a dropdown populated from the actual `.yml/.yaml` files in the selected repository, instead of requiring users to type the exact path.                                                                                                             
- New endpoint `GET /api/project/{project_id}/repositories/{repository_id}/playbooks?branch=<optional>` — local/file repos read straight off disk, git/ssh/https repos get a scratch checkout (one per branch, since `git clone` is single-branch) walked for playbook files.                                        
- Frontend picker mirrors the existing git-branch autocomplete pattern exactly, falling back to free text if the list can't be loaded (bad credentials, unreachable repo, etc).

### DIFF
--- a/api/projects/repository.go
+++ b/api/projects/repository.go
@@ -95,9 +95,13 @@ func (c *RepositoryController) GetRepositoryPlaybooks(w http.ResponseWriter, r *
 			branch = repo.GitBranch
 		}
 
+		if err := db.ValidateGitBranch(branch, "repository"); err != nil {
+			helpers.WriteError(w, err)
+			return
+		}
+
 		repoCopy := repo
 		repoCopy.GitBranch = branch
-
 		// Clone() does a single-branch clone (git clone --branch <branch>), so a
 		// scratch checkout can only ever serve the branch it was first cloned
 		// with. Key the scratch dir by branch (hashed, since branch names can
--- a/api/projects/repository.go
+++ b/api/projects/repository.go
@@ -1,15 +1,35 @@
 package projects
 
 import (
+	"crypto/sha1"
 	"errors"
 	"fmt"
 	"net/http"
+	"os/exec"
+	"time"
 
 	"github.com/semaphoreui/semaphore/api/helpers"
 	"github.com/semaphoreui/semaphore/db"
 	"github.com/semaphoreui/semaphore/db_lib"
+	"github.com/semaphoreui/semaphore/pkg/task_logger"
 	"github.com/semaphoreui/semaphore/util"
 )
+
+// nopLogger is a no-op task_logger.Logger used for git operations triggered
+// outside of a task run (e.g. browsing repository files for the playbook
+// picker), where there is no task log to write to.
+type nopLogger struct{}
+
+func (nopLogger) Log(string)                                   {}
+func (nopLogger) Logf(string, ...any)                          {}
+func (nopLogger) LogWithTime(time.Time, string)                {}
+func (nopLogger) LogfWithTime(time.Time, string, ...any)       {}
+func (nopLogger) LogCmd(*exec.Cmd)                             {}
+func (nopLogger) SetStatus(task_logger.TaskStatus)             {}
+func (nopLogger) AddStatusListener(task_logger.StatusListener) {}
+func (nopLogger) AddLogListener(task_logger.LogListener)       {}
+func (nopLogger) SetCommit(string, string)                     {}
+func (nopLogger) WaitLog()                                     {}
 
 // RepositoryMiddleware ensures a repository exists and loads it to the context
 func RepositoryMiddleware(next http.Handler) http.Handler {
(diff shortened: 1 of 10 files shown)


api/projects/repository.go
```diff
@@ -1,16 +1,36 @@
...
 	"github.com/semaphoreui/semaphore/db"
 	"github.com/semaphoreui/semaphore/db_lib"
+	"github.com/semaphoreui/semaphore/pkg/task_logger"
 	"github.com/semaphoreui/semaphore/util"
 )
 
+// nopLogger is a no-op task_logger.Logger used for git operations triggered
+// outside of a task run (e.g. browsing repository files for the playbook
+// picker), where there is no task log to write to.
+type nopLogger struct{}
```
wahyudibo: There's a duplicate of `nopLogger` struct implementation in https://github.com/semaphoreui/semaphore/blob/75c989804fceb777df66d6f323f2a94a9be64d92/db_lib/CmdGitClient_injection_test.go#L26. I suggest moving the definition under `pkg/task_logger/task_logger.go` and reference it both in here and `CmdGitClient_injection_test.go`

web/src/components/TemplateForm.vue
```diff
@@ -611,19 +630,23 @@ export default {
...
       this.setBranch = false;
+      this.loadPlaybooks();
     },
 
     async repositoryId() {
       this.branches = null;
+      this.playbooks = null;
 
       await this.loadBranches();
+      await this.loadPlaybooks();
```
wahyudibo: i think `this.loadBranches()` and `this.loadPlaybooks();` can run in parallel instead of sequentially. Should we implement something like `await Promise.all([this.loadBranches(), this.loadPlaybooks()])`?

web/src/components/TemplateForm.vue
```diff
@@ -646,6 +669,7 @@ export default {
 
   async created() {
     await this.loadBranches();
+    await this.loadPlaybooks();
```
wahyudibo: same comment as https://github.com/semaphoreui/semaphore/pull/4007/changes#r3530520756

api/projects/repository.go
```diff
@@ -76,6 +96,64 @@ func (c *RepositoryController) GetRepositoryBranches(w http.ResponseWriter, r *h
...
+		git := db_lib.GitRepository{
+			Repository: repoCopy,
+			TmpDirName: fmt.Sprintf("repository_%d_browse_%x", repo.ID, branchHash[:4]),
+			Client:     db_lib.CreateDefaultGitClient(c.keyInstaller),
+			Logger:     nopLogger{},
+		}
+
+		var err error
+		if err = git.ValidateRepo(); err != nil {
+			err = git.Clone()
```
wahyudibo, line 120: a broken `git.Clone()` process can leave an orphaned directory with a corrupt `.git` directory. Since `git.ValidateRepo()` only validates whether the directory exists, not the validity of the git repo, it can leave the repo in a stuck state. The scenario will be something like this:

1. `git.ValidateRepo()` -> err is not nil because the directory not exists -> git.Clone() failed -> directory already created but corrupted due to various reasons (failed auth, transient network error, etc)
2. next attempt, directory already present, so it passes `git.ValidateRepo()` -> `git.Pull()` -> failed due to corrupted git state

my suggestion is to delete the directory after a failed `git.Clone()`, so the subsequent attempt will always start fresh. Error in `git.Pull()` doesn't require this treatment though because the cloning process already succeeded, which means the git directory is valid

wahyudibo:
Hi @rici4kubicek @fiftin i found a bug, not a part of this PR but it's kinda related. When i add a repository with type `local`, the `loadBranches()` function returns `400` and it blocks `loadPlaybooks()` api call, thus making the playbook dropdown not load the list of playbooks. Looking at the `loadBranches()` function, it turns out the function lacks a `try catch` exception, unlike `loadPlaybooks()` which is causing this issue. After i changed the code from this

```
async loadBranches() {
      if (this.repositoryId == null) {
        return;
      }

      this.branches = await this.loadProjectEndpoint(`/repositories/${this.repositoryId}/branches`);
    },
```

to this

```
async loadBranches() {
      if (this.repositoryId == null) {
        return;
      }

      try {
        this.branches = await this.loadProjectEndpoint(`/repositories/${this.repositoryId}/branches`);
      } catch (e) {
        this.branches = null;
      }
    },
```

The dropdown appears for the playbooks, as shown in the image below. 



It would be great if we can also include the fix in this PR.

api/projects/repository.go
rici4kubicek: Good catch. Moved to `task_logger.NopLogger` in `pkg/task_logger/task_logger.go`, now shared by `api/projects/repository.go` and `db_lib/CmdGitClient_injection_test.go`. Fixed in dc5392a6.

web/src/components/TemplateForm.vue
rici4kubicek: Agreed, no dependency between them. Switched to `await Promise.all([this.loadBranches(), this.loadPlaybooks()])`. Fixed in dc5392a6.

web/src/components/TemplateForm.vue
rici4kubicek: Same fix applied here too — `Promise.all`. Fixed in dc5392a6.

api/projects/repository.go
rici4kubicek: Right, `ValidateRepo()` only checks dir existence not git validity. `GitRepository.Clone()` now does `os.RemoveAll` on the target path when `Client.Clone()` fails, so the next attempt starts fresh. `Pull()` left untouched per your note — a successful prior clone means the git state is trustworthy. Fixed in dc5392a6.

rici4kubicek:
@wahyudibo Thanks for catching this — added try/catch to `loadBranches()` matching `loadPlaybooks()`'s pattern, so a 400 (e.g. local repo type) no longer blocks the playbook list from loading. Fixed in 11a9cd70.

fiftin:
Hi @rici4kubicek it is really-really useful PR. Thank you for it!

PR #4182, opened 2026-09-01
fix(docker): pin bcrypt below 5 so passlib can hash

### LINKED ISSUE #4180: Problem: password_hash('bcrypt') fails in Docker image — unpinned passlib + bcrypt 5.0.0
### Issue

`password_hash('bcrypt')` is broken in the official Docker image. Any use of the filter fails, regardless of the password supplied:

```
The filter plugin 'ansible.builtin.password_hash' failed: Could not hash the secret:
password cannot be longer than 72 bytes, truncate manually if necessary (e.g. my_password[:72])
```

The "72 bytes" message is misleading — it is **not** about the user's password. It fails with a 4-byte password. The >72-byte value is passlib's *own* internal self-test.

**Root cause:** the bundled ansible venv contains `passlib 1.7.4` + `bcrypt 5.0.0`, which are incompatible.

During backend initialisation, `passlib/handlers/bcrypt.py::_finalize_backend_mixin` calls `detect_wrap_bug(IDENT_2A)` (line 421), a BSD-wraparound self-test that hashes a deliberately oversized secret:

```python
secret = (b"0123456789"*26)[:255]   # 255 bytes
```

bcrypt <5 silently truncated that to 72 bytes. **bcrypt 5.0.0 raises `ValueError` instead**, so passlib's bcrypt backend never finishes loading and every subsequent call to the filter fails.

Only bcrypt is affected — other schemes work fine:

| scheme | result |
|---|---|
| `sha512` | ✅ works |
| `sha256` | ✅ works |
| `bcrypt` | ❌ fails |

**Why the image is affected.** In `deployment/docker/server/Dockerfile`:

```dockerfile
pip3 install --upgrade pip ansible==${ANSIBLE_VERSION} boto3 botocore requests pywinrm passlib paramiko && \
```

`passlib` is unpinned, and `bcrypt` is not listed at all — it is pulled in **transitively via paramiko**:

```
Name: bcrypt
Version: 5.0.0
Required-by: paramiko
```

So each rebuild resolves `bcrypt` to the newest release. Since passlib 1.7.4 (Oct 2020) is the final passlib release, this combination will keep reproducing on every future image build unless bcrypt is constrained.

This is a regression from the user's point of view: the same playbooks ran green on `v2.18.27` and fail on `v2.19.12`.

### Reproduction

```yaml
- hosts: localhost
  connection: local
  gather_facts: false
  tasks:
    - debug:
        msg: "{{ 'test' | password_hash('bcrypt') }}"
```

Fails inside the image with the error above. `password_hash('sha512')` on the same run succeeds.

### Impact

Ansible (task execution), Docker

### Installation method

Docker

### Database

SQLite

### Semaphore Version

```
v2.19.12-012ed06-1788086369
```

### Ansible Version

```bash
ansible 13.5.0
ansible-core 2.20.8
bcrypt 5.0.0
passlib 1.7.4
cryptography 50.0.1
paramiko 5.0.0
# venv: /opt/semaphore/apps/ansible/13.5.0/venv
```

### Additional information

**Upstream context**

- ansible-core [#85919](https://github.com/ansible/ansible/issues/85919) — same failure. Maintainer guidance there is explicit that, since `bcrypt` is a dependency of an *optional* dependency, constraining it is the responsibility of whoever installs passlib:
  > `pip install passlib 'bcrypt<5'`
- ansible-core [#85970](https://github.com/ansible/ansible/pull/85970) (merged 2025-10-21) adds a `crypt`/`libxcrypt` backend that bypasses passlib entirely. This landed in **ansible-core 2.21**, so the image's 2.20.8 predates it — I confirmed `ansible/utils/encrypt.py` in 2.20.8 has no libxcrypt path, while 2.21.3 does.
- prometheus-community/ansible hit the identical error ([#660](https://github.com/prometheus-community/ansible/issues/660)) and resolved it by pinning `bcrypt<5` ([#722](https://github.com/prometheus-community/ansible/pull/722)), explicitly declining the bcrypt v5 bump ([#723](https://github.com/prometheus-community/ansible/pull/723)).

official-burak:
Fixes #4180.

bcrypt 5 makes passlib 1.7.4 fail its wraparound self-test, so `password_hash('bcrypt')` errors in the Docker image even for short passwords. This pins `bcrypt>=4,<5` next to passlib in the server, runner, and job images, matching the ansible-core advice.

Confirmed locally: passlib 1.7.4 + bcrypt 5.0.0 raises the 72-byte error; bcrypt 4.x hashes fine.

### DIFF
--- a/deployment/docker/job/Dockerfile
+++ b/deployment/docker/job/Dockerfile
@@ -21,13 +21,14 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
         unzip \
     && rm -rf /var/lib/apt/lists/*
 
-
+# passlib 1.7.4 cannot load its bcrypt backend on bcrypt 5+ (#4180)
 RUN pip3 install --no-cache-dir --break-system-packages \
         "ansible-core>=2.16,<2.18" \
         "ansible==${ANSIBLE_VERSION}" \
         jmespath \
         netaddr \
-        passlib
+        passlib \
+        'bcrypt>=4,<5'
 
 RUN set -eux; \
     for bin in sh cat chmod rm ls echo ssh ssh-agent ssh-add setsid git ansible-playbook python3; do \
--- a/deployment/docker/runner/Dockerfile
+++ b/deployment/docker/runner/Dockerfile
@@ -85,11 +85,12 @@ ARG ANSIBLE_VERSION=13.5.0
 ENV ANSIBLE_VERSION=${ANSIBLE_VERSION}
 ARG ANSIBLE_VENV_PATH=/opt/semaphore/apps/ansible/${ANSIBLE_VERSION}/venv
 
+# passlib 1.7.4 cannot load its bcrypt backend on bcrypt 5+ (#4180)
 RUN apk add --no-cache -U python3-dev build-base openssl-dev libffi-dev cargo && \
      mkdir -p ${ANSIBLE_VENV_PATH} && \
      python3 -m venv ${ANSIBLE_VENV_PATH} --system-site-packages && \
      source ${ANSIBLE_VENV_PATH}/bin/activate && \
-     pip3 install --upgrade pip ansible==${ANSIBLE_VERSION} boto3 botocore requests pywinrm passlib && \
+     pip3 install --upgrade pip ansible==${ANSIBLE_VERSION} boto3 botocore requests pywinrm passlib 'bcrypt>=4,<5' && \
      apk del python3-dev build-base openssl-dev libffi-dev cargo && \
      rm -rf /var/cache/apk/* && \
      find ${ANSIBLE_VENV_PATH} -iname __pycache__ | xargs rm -rf && \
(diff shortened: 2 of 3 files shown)


fiftin:
Hi @official-burak thank you for your contribution.

But we need some test for this case. I will think how we can do this.

UPDATE:

Could you provide a playbook to reproduce the issue?

official-burak:
Sure. This is enough to hit it inside the current Docker ansible venv:

```yaml
- hosts: localhost
  connection: local
  gather_facts: false
  tasks:
    - debug:
        msg: "{{ 'test' | password_hash('bcrypt') }}"
```

With passlib 1.7.4 + bcrypt 5 you get the 72-byte wraparound error on that four-byte password. `password_hash('sha512')` on the same run is fine. Totally fair on wanting a test for this, happy to help once you've got a shape in mind.

PR #4250, opened 2026-09-18
fix: pass Ansible secret extra vars as JSON

### LINKED ISSUE #4249: Problem: Ansible secret extra-vars are parsed incorrectly when values contain spaces
### Issue

Variable Group secret variables are passed to Ansible differently from regular extra variables.

Regular extra variables are serialized into a JSON `--extra-vars` argument, but secret variables are currently appended separately as:

    --extra-vars name=value

This causes values containing spaces or other complex content to be parsed incorrectly by Ansible.

For example, a secret variable containing:

    value with spaces

is not transported in the same safe way as the existing JSON extra-vars payload.

This is especially problematic for values such as SSH public keys, tokens containing spaces, and other structured secret values.

### Current behavior

Semaphore builds the normal extra-vars JSON payload first, then appends each Environment Secret Variable as another CLI argument:

    --extra-vars name=value

### Expected behavior

Environment Secret Variables of type `var` should be merged into the existing JSON extra-vars map before serialization.

This preserves the exact value and also keeps the existing secret-variable override precedence.

Environment secrets of type `env` should remain environment variables and should not be included in the Ansible JSON extra-vars payload.

### Reproduction

1. Create a Variable Group.
2. Add a Secret Variable.
3. Set its value to a string containing spaces.
4. Attach the Variable Group to an Ansible template.
5. Run a playbook that validates the complete variable value.

The value is not transported correctly when passed using `name=value`.

### Tested fix

I tested a change that merges `EnvironmentSecretVar` values into `getEnvironmentExtraVarsJSON()` and removes the separate `--extra-vars name=value` arguments.

A regression test verifies:

- values containing spaces;
- multiline values;
- secret variable precedence over normal extra vars;
- exclusion of EnvironmentSecretEnv values.

The fix has also been verified end-to-end with an Ansible task.

### Semaphore version

Originally reproduced on v2.19.7.

The same separate `name=value` transport is still present in the current `develop` branch.

Iceman-ZP:
## What does this PR do?

Fixes Ansible Variable Group secret variables being passed as separate `--extra-vars name=value` arguments.

Secret variables of type `var` are now merged into the existing JSON extra-vars payload before it is serialized and passed to Ansible.

This preserves values containing spaces and other complex content.

Environment secrets of type `env` are unchanged and continue to be passed as environment variables.

## Why?

Regular extra variables are already passed to Ansible as JSON, but Variable Group secret variables were appended separately using:

    --extra-vars name=value

This can cause Ansible to parse values containing spaces incorrectly.

Fixes #4249.

## Behavior

The change preserves:

- secret variable precedence over regular extra variables;
- EnvironmentSecretEnv behavior;
- a single JSON `--extra-vars` payload.

## Tests

Added a regression test covering:

- secret values containing spaces;
- multiline secret values;
- secret variable override precedence;
- exclusion of EnvironmentSecretEnv values;
- use of a single JSON extra-vars payload.

Tested with:

    go test ./services/tasks -run TestGetPlaybookArgs_EnvironmentSecretVarsUseJSON -count=1

and:

    go test ./services/tasks -count=1

Both pass on the current `develop` branch.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of database secret variables during task execution.
  * Secret values, including those containing spaces or multiple lines, are now passed reliably as part of the consolidated variable payload.
  * Secret variables correctly take precedence over plain variables with the same name.
  * Secrets intended for environment-level configuration are excluded from task variables.


fiftin:
@Iceman-ZP as I remember the `env` is stored in a database in the `task` table. 

Secrets shouldn't be stored in the database.

Iceman-ZP:
Thanks for pointing this out.

Just to clarify, my intention here wasn't to change how Semaphore stores
secrets.

I ran into this while using a multiline private key as an Ansible secret
variable. With the current --extra-vars name=value form, part of the value
was lost during parsing. Passing the same value as JSON fixed the issue in
my setup, so I opened this PR mainly to share the problem and the working
fix I used.

If this approach conflicts with how secrets are expected to be handled
internally, I'm happy to leave the implementation decision to the
maintainers.

вс, 20 сент. 2026 г. в 09:46, Denis Gukov ***@***.***>:

PR #4257, opened 2026-09-21
fix(tasks): pass secret extra-vars as JSON to preserve spaces

### LINKED ISSUE #4249: Problem: Ansible secret extra-vars are parsed incorrectly when values contain spaces

EyJunge1:
## Summary
- Merges Variable Group secrets of type `var` into the shared extra-vars map so Ansible receives them inside the JSON `--extra-vars` payload instead of fragile `name=value` CLI args.
- Removes the duplicate secret loops for Ansible, Shell, and Terraform (secrets already come from `getEnvironmentExtraVars`).
- Secrets of type `env` stay process environment only.

Closes #4249

## Test plan
- [x] `go test ./services/tasks/ -run 'ExtraVars|ShellArgs|TerraformArgs|PlaybookArgs_Secret' -count=1`
- [ ] Optional: run an Ansible task with a secret value containing spaces and confirm the full string arrives

## Summary by CodeRabbit

- **Bug Fixes**
  - Environment secret variables are now consistently passed to shell, Terraform, and Ansible tasks through their standard variable inputs.
  - Secret values containing spaces or line breaks are preserved correctly.
  - Secret variables take precedence over duplicate values defined in environment settings or survey responses.


fiftin:
@EyJunge1 Semaphore stores `envVars` in a database have ability to see task running params. But secrets shouldn't be stored in the database.

fiftin:
I'm closing the PR, please reopen if you fix it.

EyJunge1:
@fiftin Fixed.

Secrets are no longer merged into `Environment.JSON` / Task env (DB-visible). Ansible gets them only as a separate ephemeral JSON `--extra-vars` at runtime, so values with spaces still work.

